PR #83, opened 2026-05-28
feat: API access control for web-chat and API

Calmingstorm:
## Summary

- Web-chat and API now propagate authenticated identity (tier, user_id, allowed_hosts) instead of hardcoding "web-user"
- Tier-based tool filtering enforced for web-chat and API, matching Discord behavior
- API token `allowed_hosts` field now enforced via transient host scoping
- New `ApiTokenManager` with CRUD + hot-reload (stored in `data/api_tokens.json`)
- REST endpoints: `GET/POST /api/tokens`, `PUT/DELETE /api/tokens/{user_id}`, `POST /api/tokens/{user_id}/regenerate`
- WebUI: System > API Tokens page — create, edit, delete, regenerate tokens with full permission config
- WebSocket chat hardened: 10 msg/min rate limit, 300s timeout for long-running tools
- Legacy single `api_token` continues working as admin (backward compatible)
- Login with legacy token now binds admin identity to session (fixes web-chat falling back to default policy)

## Test plan

- [ ] Log into WebUI, open web-chat, verify identity shows as `api-admin` (not `web-user`) in logs
- [ ] Create a `user` tier API token from the new UI page, test that it only gets read-only tools via `/api/execute`
- [ ] Create a token with `allowed_hosts: ["playground"]`, verify commands on `localhost` are denied
- [ ] Regenerate a token, verify the old one stops working and the new one works
- [ ] Delete a token, verify it's gone
- [ ] Verify legacy single `api_token` in config.yml still works as admin
- [ ] Spam web-chat messages, verify rate limit kicks in after 10/min
- [ ] Run existing test suite — 0 new failures (27 pre-existing failures unrelated to this change)

### DIFF
--- a/src/discord/client.py
+++ b/src/discord/client.py
@@ -420,6 +420,9 @@ class OdinBot(commands.Bot):
             overrides_path=config.permissions.overrides_path,
         )
 
+        from ..permissions.token_manager import ApiTokenManager
+        self.api_token_manager = ApiTokenManager("./data/api_tokens.json")
+
         self.tool_memory = ToolMemory("./data/tool_memory.json")
 
         # Wire optional services into skill manager for expanded skill context
--- a/src/health/server.py
+++ b/src/health/server.py
@@ -168,8 +168,11 @@ def _make_auth_middleware(
             if token and hmac.compare_digest(bearer_value, token):
                 request._session_id = "admin-token"
                 return await handler(request)
-            # Check multi-token identities
-            identity = web_config.resolve_api_identity(bearer_value) if hasattr(web_config, "resolve_api_identity") else None
+            # Check dynamic token manager first, then static config tokens
+            tm = request.app.get("token_manager")
+            identity = tm.resolve(bearer_value) if tm else None
+            if identity is None and hasattr(web_config, "resolve_api_identity"):
+                identity = web_config.resolve_api_identity(bearer_value)
             if identity is not None:
                 request._session_id = identity.user_id
                 request._api_identity = identity
@@ -188,12 +191,14 @@ def _make_auth_middleware(
             if token and hmac.compare_digest(query_token, token):
                 request._session_id = "admin-token"
                 return await handler(request)
-            if hasattr(web_config, "resolve_api_identity"):
+            tm = request.app.get("token_manager")
+            identity = tm.resolve(query_token) if tm else None
+            if identity is None and hasattr(web_config, "resolve_api_identity"):
                 identity = web_config.resolve_api_identity(query_token)
-                if identity is not None:
-                    request._session_id = identity.user_id
-                    request._api_identity = identity
-                    return await handler(request)
+            if identity is not None:
+                request._session_id = identity.user_id
+                request._api_identity = identity
+                return await handler(request)
             if session_manager.validate(query_token):
                 request._session_id = query_token
                 session_identity = session_manager.get_identity(query_token)
@@ -500,8 +505,10 @@ class HealthServer:
         from ..web.api import setup_api
         from ..web.websocket import setup_websocket
         setup_api(self._app, bot)
+        self._app["token_manager"] = getattr(bot, "api_token_manager", None)
         self._ws_manager = setup_websocket(
             self._app, bot, api_token=self._web_config.api_token,
+            web_config=self._web_config,
         )
         # Wire audit events to WebSocket for live dashboard/log updates
         ws_mgr = self._ws_manager
--- a/src/permissions/host_access.py
+++ b/src/permissions/host_access.py
@@ -40,6 +40,7 @@ class HostAccessManager:
         self._users: dict[str, HostAccessEntry] = {}
         self._default_policy = HostAccessEntry()
         self._available_hosts: list[str] = available_hosts or []
+        self._transient_scopes: dict[str, list[str]] = {}
         self._load()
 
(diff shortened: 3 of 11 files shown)


Calmingstorm:
## Review: PR #83 — API access control for web-chat and API

I found security/correctness issues that should block this as-is.

### Blocking findings

1. **Dynamic API tokens do not actually enable auth in a config with only dynamic tokens**

`_make_auth_middleware()` and `/api/auth/login` decide whether auth is configured from only `web_config.api_token` and `web_config.api_tokens`:

- `src/health/server.py:156-160`
- `src/web/api.py:157-160`

The newly added token manager lives in `app["token_manager"]` / `bot.api_token_manager`, but it is not included in `has_any_token`. If an install relies only on `data/api_tokens.json` with no legacy/static token configured, auth middleware treats the API as dev/no-auth and permits every `/api/*` request. `/api/auth/login` also reports “invalid token” because it bails on the same static-only check path.

This defeats the dynamic-token feature in the exact deployment mode the UI encourages. The middleware should consider the dynamic token manager’s loaded token count/existence when deciding auth mode, or dynamic tokens should be disabled unless a legacy/static token also exists. The safer default is: if the token manager exists and contains any token, auth is configured.

2. **Transient tier/host scopes are not concurrency-safe and can cross-contaminate requests**

`process_web_chat()` writes request-scoped privileges into global shared managers keyed only by `user_id`:

- `src/web/chat.py:165-178`
- `src/permissions/manager.py:64-76`
- `src/permissions/host_access.py:84-100,110-119`

This is unsafe under concurrent web/API/WebSocket requests for the same token identity. Example: two `/api/execute` calls for `user_id="ci"` run concurrently, one with a host-limited token and one with broader/elevated privileges, or one finishes and clears the key while the other is still executing. Tool permission checks and host resolution happen later during async tool execution, so the effective tier/host scope can be whichever request last wrote the shared dict, or no scope at all after a sibling request clears it.

This is not just theoretical; the design uses process-global mutable state and `await` spans the whole LLM/tool loop. The transient tier/scope pattern needs to be request-local, e.g. pass an `AccessContext` through `ToolExecutor.execute()` / permission checks, use `contextvars`, or key scopes by a unique request id and propagate that id to all permission/host checks. Keying only by user id is the flaw.

3. **`ToolExecutor` has an existing shared `_current_user_id` state that makes host scoping worse under parallel/concurrent tools**

Host checks depend on `self._current_user_id`, which is overwritten per `execute()` call:

- `src/tools/executor.py:127-135,164-171`
- multi-host path: `src/tools/executor.py:606-618`

Because one `ToolExecutor` instance is shared by the bot, concurrent API requests or parallel tool calls can overwrite `_current_user_id` while another tool is still resolving hosts. The PR’s transient host scope relies on that mutable executor state too, so even if the transient manager dict were fixed, host enforcement can still race unless the user/access context is kept local to the execution path instead of stored on the executor object.

4. **WebSocket identity resolution omits dynamic tokens unless middleware has already populated `_api_identity`**

`WebSocketManager._resolve_identity()` checks sessions and static config tokens only:

- `src/web/websocket.py:56-67`

The middleware path may set `_api_identity` for `?token=...`, but the WebSocket handler also performs its own token validation/resolution (`src/web/websocket.py:71-83`). That local resolver should include the dynamic token manager too, otherwise future refactors, tests, or non-middleware construction will silently regress to static-only identity. Right now this is brittle duplicated auth logic.

### Non-blocking but important hardening

- **Token records are stored as raw bearer tokens** in `data/api_tokens.json` (`src/permissions/token_manager.py`). The file is under `data/` and gitignored, which helps, but the safer pattern is to store a token hash plus prefix/label and only show the raw token once at creation/regeneration. Dynamic UI-created API tokens are high-value secrets.
- **Token creation/update accepts `allowed_tools` and `allowed_hosts` without type/known-value validation** (`src/web/api.py:3036-3044,3070-3075`). Bad types can persist malformed config or cause runtime errors/odd policy behavior. Validate arrays of strings, tier enum, host names against `available_hosts`, and tool names against registry.
- **Regression tests are too thin for the risky parts.** `tests/test_execute_api.py` includes a mirrored fake handler, and the real-handler tests don’t exercise auth middleware, dynamic-token-only mode, WebSocket dynamic tokens, or concurrent same-user request scoping. Add tests that prove two simultaneous requests with the same `user_id` but different scopes cannot affect each other.

### Validation I ran

Observed facts:

```text
python3 -m py_compile src/permissions/token_manager.py src/permissions/manager.py src/permissions/host_access.py src/web/api.py src/web/chat.py src/web/websocket.py src/health/server.py
pytest -q tests/test_execute_api.py
pytest -q tests/test_web_auth_policy.py tests/test_permissions.py
git diff --check master...HEAD
node --check ui/js/pages/api-tokens.js
node --check ui/js/pages/system.js
```

Results:

```text
tests/test_execute_api.py: 15 passed
tests/test_web_auth_policy.py tests/test_permissions.py: 8 passed
git diff --check: clean
node --check changed UI files: clean
```

Judgment: the PR is directionally useful, but **not safe to merge yet**. The transient tier/scope approach is the core architectural risk; for orchestrator-style concurrent requests, access control needs to be request-local, not patched into shared global managers for the duration of an async loop.

Calmingstorm:
Closing in favor of new PR with Odin review fixes applied.

PR #85, opened 2026-05-28
feat: API access control for web-chat and API (v3)

Calmingstorm:
## Summary

Replaces #84, incorporating round 2 of Odin's review findings.

**Round 2 fixes:**
- WebSocket auth now resolves dynamic tokens before the reject gate — fixes mixed legacy+dynamic deployments where dynamic tokens were silently rejected
- Token delete/regenerate destroys all sessions for that user_id via `SessionManager.destroy_by_user_id()` — revoked tokens can no longer linger through existing sessions
- Token file loader validates tier enum, allowed_tools/allowed_hosts types — malformed entries logged and skipped, not silently loaded as dangerous identities

**Previous fixes (from round 1):**
- Concurrency-safe scoping via contextvars
- Token hashing (SHA-256) at rest
- Dynamic token manager in auth detection
- Input validation on API endpoints

## Test plan

- [ ] Dynamic token via WebSocket with legacy api_token also configured — should connect and get correct identity
- [ ] Delete a token, verify existing sessions for that user_id are destroyed
- [ ] Regenerate a token, verify old sessions destroyed and old token stops working
- [ ] Malformed `data/api_tokens.json` entry (bad tier, non-list tools) — should skip with warning log
- [ ] All 65 relevant tests pass, 0 new failures

### DIFF
--- a/src/discord/client.py
+++ b/src/discord/client.py
@@ -420,6 +420,9 @@ class OdinBot(commands.Bot):
             overrides_path=config.permissions.overrides_path,
         )
 
+        from ..permissions.token_manager import ApiTokenManager
+        self.api_token_manager = ApiTokenManager("./data/api_tokens.json")
+
         self.tool_memory = ToolMemory("./data/tool_memory.json")
 
         # Wire optional services into skill manager for expanded skill context
--- a/src/health/server.py
+++ b/src/health/server.py
@@ -168,8 +168,11 @@ def _make_auth_middleware(
             if token and hmac.compare_digest(bearer_value, token):
                 request._session_id = "admin-token"
                 return await handler(request)
-            # Check multi-token identities
-            identity = web_config.resolve_api_identity(bearer_value) if hasattr(web_config, "resolve_api_identity") else None
+            # Check dynamic token manager first, then static config tokens
+            tm = request.app.get("token_manager")
+            identity = tm.resolve(bearer_value) if tm else None
+            if identity is None and hasattr(web_config, "resolve_api_identity"):
+                identity = web_config.resolve_api_identity(bearer_value)
             if identity is not None:
                 request._session_id = identity.user_id
                 request._api_identity = identity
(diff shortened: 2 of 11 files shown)


Calmingstorm:
## Review: PR #85 API access control v3

I verified the round-2 fixes directly against the diff and with targeted smoke tests. The contextvars tier/host scoping is concurrency-isolated, token storage no longer writes raw tokens, malformed token-file entries are skipped, dynamic WebSocket token resolution works in a mixed legacy+dynamic setup, and `destroy_by_user_id()` removes session entries for deleted/regenerated token identities.

One blocking revocation gap remains:

### Blocking: deleting/regenerating a token does not revoke already-open WebSocket connections

`DELETE /api/tokens/{user_id}` and `/regenerate` call `SessionManager.destroy_by_user_id()` (`src/web/api.py:3120-3123`, `3137-3139`), but WebSocket connections are tracked only in `WebSocketManager._clients`. When a WebSocket is authenticated with a dynamic raw token, `handle()` stores the resolved identity on the socket (`ws._odin_identity = identity`, `src/web/websocket.py:97-101`) and `_handle_chat()` keeps trusting that cached identity for every later message (`src/web/websocket.py:193-206`). There is no per-message revalidation and no hook from token delete/regenerate to close live sockets for that `user_id`.

This means a client can connect with a token, the token can be deleted or regenerated, and the existing WebSocket can continue executing chat/tool requests with the old identity until it disconnects. The same applies to WebSockets opened with a session id: deleting the session from `SessionManager` does not close the already-prepared socket, and `_handle_chat()` does not check whether `ws._odin_session_id` is still valid.

That violates the PR’s stated revocation guarantee: “old token stops working immediately.” It only stops new HTTP/session auth, not already-established WebSocket command channels.

Recommended fix: add a WebSocket revocation path and/or per-message validation. For example:

- add `WebSocketManager.close_by_user_id(user_id)` that closes/removes sockets whose `_odin_identity.user_id` matches;
- expose the manager on `app` so token delete/regenerate can call it after `destroy_by_user_id()`;
- for session-authenticated sockets, either close sockets whose `_odin_session_id` was destroyed or revalidate session ids before each chat message;
- add tests proving a deleted/regenerated token cannot continue sending WebSocket chat over an existing connection.

### Validation run

Targeted tests mostly pass, with the known non-security contract mismatch still present:

```text
python3 -m pytest tests/test_execute_api.py tests/test_web_chat.py tests/test_websocket_handler.py tests/test_web_auth_policy.py tests/test_permissions.py tests/test_web_api_new_endpoints.py tests/test_health_endpoints.py -q

1 failed, 84 passed
FAILED tests/test_web_chat.py::TestProcessWebChat::test_exception_handling
AssertionError: assert 'Error processing' in 'Something went wrong processing your message. Check server logs for details.'
```

Additional smoke checks passed:

- raw dynamic tokens are not persisted to `data/api_tokens.json`;
- old token stops resolving after regeneration;
- new token stops resolving after delete;
- malformed loader entries with invalid tier/tools/hosts are skipped;
- contextvars tier/host scopes remained isolated across concurrent web-chat calls;
- dynamic token identity resolves through the WebSocket resolver while a legacy token is also configured;
- `SessionManager.destroy_by_user_id()` removes matching session ids.

Verdict: close on the original findings, but not safe to merge while live WebSocket revocation remains open. The old gods call this “almost fixed,” which is the most dangerous shape of fixed.

Calmingstorm:
Closing for v4 with WebSocket revocation fix.

PR #86, opened 2026-05-28
feat: API access control for web-chat and API (v4)

Calmingstorm:
## Summary

Replaces #85. Adds WebSocket connection revocation — the last blocking finding.

**Round 3 fix:**
- `WebSocketManager.close_by_user_id()` closes all live WebSocket connections for a revoked user_id (code 4002, "token revoked")
- Called from both token delete and regenerate endpoints alongside `SessionManager.destroy_by_user_id()`
- All three access paths now fully revoked: HTTP sessions, WebSocket connections, raw bearer tokens

**Full change set across 4 commits:**
1. Identity propagation, tier enforcement, host scoping, token manager, CRUD endpoints, WebUI page, rate limiting, timeout
2. Concurrency-safe contextvars, token hashing, auth detection, input validation
3. WS dynamic token auth, session revocation, loader validation
4. WebSocket connection revocation

## Test plan
- [ ] Connect WebSocket with dynamic token, delete token, verify WS is closed with code 4002
- [ ] Regenerate token, verify existing WS connections closed
- [ ] All 65 relevant tests pass

### DIFF
--- a/src/discord/client.py
+++ b/src/discord/client.py
@@ -420,6 +420,9 @@ class OdinBot(commands.Bot):
             overrides_path=config.permissions.overrides_path,
         )
 
+        from ..permissions.token_manager import ApiTokenManager
+        self.api_token_manager = ApiTokenManager("./data/api_tokens.json")
+
         self.tool_memory = ToolMemory("./data/tool_memory.json")
 
         # Wire optional services into skill manager for expanded skill context
--- a/src/health/server.py
+++ b/src/health/server.py
@@ -522,6 +522,7 @@ class HealthServer:
             self._app, bot, api_token=self._web_config.api_token,
             web_config=self._web_config,
         )
+        self._app["ws_manager"] = self._ws_manager
         # Wire audit events to WebSocket for live dashboard/log updates
         ws_mgr = self._ws_manager
         bot.audit.set_event_callback(ws_mgr.broadcast_event)
--- a/src/health/server.py
+++ b/src/health/server.py
@@ -168,8 +168,11 @@ def _make_auth_middleware(
             if token and hmac.compare_digest(bearer_value, token):
                 request._session_id = "admin-token"
                 return await handler(request)
-            # Check multi-token identities
-            identity = web_config.resolve_api_identity(bearer_value) if hasattr(web_config, "resolve_api_identity") else None
+            # Check dynamic token manager first, then static config tokens
+            tm = request.app.get("token_manager")
+            identity = tm.resolve(bearer_value) if tm else None
+            if identity is None and hasattr(web_config, "resolve_api_identity"):
+                identity = web_config.resolve_api_identity(bearer_value)
             if identity is not None:
                 request._session_id = identity.user_id
                 request._api_identity = identity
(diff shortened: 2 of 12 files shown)


Calmingstorm:
## Review: PR #86 — API access control for web-chat and API (v4)

Thanks for the v4 update. I re-reviewed the new WebSocket revocation path plus the surrounding auth/session/token flow, and also reran the targeted web/API/WebSocket tests.

### What I verified

- `WebSocketManager.close_by_user_id()` now walks live `_clients`, finds sockets with `_odin_identity.user_id`, closes them with code `4002`, and removes them from client/subscriber sets.
- Dynamic-token delete and regenerate both now call `SessionManager.destroy_by_user_id()` and `WebSocketManager.close_by_user_id()`.
- WebSocket handshake resolves dynamic tokens before rejecting, so mixed legacy + dynamic token deployments are not rejected at the gate.
- Request tier and host scoping are using `contextvars` and reset in `finally`, which addresses the earlier concurrent-request cross-contamination issue for the web-chat/API request path.
- Dynamic tokens are stored hashed, raw tokens are returned only on create/regenerate, and malformed persisted entries for tier/tools/hosts are skipped.

Targeted test run:

```
python3 -m pytest tests/test_web_auth_policy.py tests/test_execute_api.py tests/test_websocket_handler.py tests/test_web_chat.py -q
# 62 passed, 1 failed
```

The remaining failure is the pre-existing web-chat exception-message contract mismatch in `TestProcessWebChat.test_exception_handling`; it is not related to token revocation. The revocation issue itself looks addressed.

### Findings

I do not see a blocking security bug in PR #86's stated revocation fix. Delete/regenerate now revoke all three intended access paths: HTTP sessions, live WebSocket connections, and future raw bearer-token use. The previous live-WebSocket gap is closed.

I do still see one **important non-blocking gap** worth fixing before this becomes a foundation for the orchestrator architecture:

1. **No direct regression coverage for live WebSocket revocation.**

   The implementation is straightforward, but this is security-critical behavior and should not rely on review memory. Add tests that prove:

   - `close_by_user_id()` closes only sockets whose `_odin_identity.user_id` matches the revoked token user.
   - closed sockets are removed from `_clients`, `_log_subscribers`, and `_event_subscribers`.
   - `/api/tokens/{user_id}/regenerate` and `DELETE /api/tokens/{user_id}` invoke both `SessionManager.destroy_by_user_id(uid)` and `ws_manager.close_by_user_id(uid)`.
   - an already-open WS authenticated via a dynamic token cannot send another chat message after delete/regenerate. An integration-style aiohttp WS test would be ideal here, because socket state is where the old bug lived.

   The absence of this test is not a reason to block the PR by itself, but this exact class of bug has now recurred through multiple rounds. Put a spear through it in CI so we do not rediscover it in production at 03:00 like mortals with amnesia.

### Notes / minor concerns

- `update_token()` can reduce a token's tier or host/tool scope, but it does not revoke existing sessions/WebSockets. If token updates are intended to take effect immediately, that should also call session/WS revocation. If updates are intentionally eventual-until-next-login/connection, document that clearly in the UI/API. Delete/regenerate are the critical revocation paths requested here, and those are fixed.
- `GET /api/auth/session` still only checks legacy bearer tokens and sessions, not dynamic bearer tokens directly. Middleware-protected routes will still authenticate dynamic bearer tokens, so this is more of a UI/status consistency issue than an access-control bypass.
- Dynamic token file loading validates types but does not validate persisted `allowed_hosts` against currently available hosts. Runtime enforcement intersects host scope with available hosts, so this does not become host escalation; it just means stale host names can persist harmlessly.

### Verdict

**Approve from a security/access-control perspective, with a strong request to add WebSocket revocation regression tests.**

The round 3 finding is fixed: existing WebSocket connections tied to a revoked dynamic token user are now actively closed on delete/regenerate, and future raw bearer-token use is rejected because the token hash is gone or replaced.

PR #88, opened 2026-05-28
fix: token allowed_hosts authoritative when no host_access entry

Calmingstorm:
## Summary

Bug fix for v3.27.0 identity-scoped sessions. When a token has `allowed_hosts` configured but the user_id has no explicit entry in `host_access.json`, the token's hosts should be the effective set — not intersected with `default_policy`.

**The bug**: Admin token configured with all hosts still got restricted to `playground` (the default policy) because the intersection of `token_hosts ∩ default_policy` narrowed it.

**The fix**: If a request-scoped host scope is set AND the user_id has no explicit host_access entry, use the scope directly as the authoritative host list. If the user HAS an explicit entry, intersection still applies (token scope narrows the entry).

## Test plan
- [ ] API call with Admin token (all hosts configured) can run commands on localhost
- [ ] Token with subset of hosts is restricted to those hosts only
- [ ] User with explicit host_access entry + token scope = intersection still works

### DIFF
--- a/src/permissions/host_access.py
+++ b/src/permissions/host_access.py
@@ -108,9 +108,15 @@ class HostAccessManager:
         return base
 
     def get_default_host(self, user_id: str) -> str:
+        scope = _request_host_scope.get()
+        has_own_entry = user_id in self._users
+        if scope is not None and not has_own_entry:
+            valid = [h for h in scope if h in self._available_hosts]
+            return valid[0] if valid else ""
         entry = self.get_entry(user_id)
         if entry.default_host and entry.default_host in self._available_hosts:
-            return entry.default_host
+            if scope is None or entry.default_host in scope:
+                return entry.default_host
         allowed = self.get_allowed_hosts(user_id)
         return allowed[0] if allowed else ""
 
--- a/src/permissions/host_access.py
+++ b/src/permissions/host_access.py
@@ -94,12 +94,15 @@ class HostAccessManager:
         _request_host_scope.reset(token)
 
     def get_allowed_hosts(self, user_id: str) -> list[str]:
+        scope = _request_host_scope.get()
+        has_own_entry = user_id in self._users
+        if scope is not None and not has_own_entry:
+            return [h for h in scope if h in self._available_hosts]
         entry = self.get_entry(user_id)
         if entry.allowed_hosts is None:
             base = list(self._available_hosts)
         else:
             base = [h for h in entry.allowed_hosts if h in self._available_hosts]
-        scope = _request_host_scope.get()
         if scope is not None:
             base = [h for h in base if h in scope]
         return base
@@ -112,12 +115,15 @@ class HostAccessManager:
         return allowed[0] if allowed else ""
 
     def is_host_allowed(self, user_id: str, host: str) -> bool:
+        scope = _request_host_scope.get()
+        has_own_entry = user_id in self._users
+        if scope is not None and not has_own_entry:
+            return host in scope and host in self._available_hosts
         entry = self.get_entry(user_id)
         if entry.allowed_hosts is None:
             allowed = host in self._available_hosts
         else:
             allowed = host in entry.allowed_hosts and host in self._available_hosts
-        scope = _request_host_scope.get()
         if scope is not None:
             allowed = allowed and host in scope
         return allowed


Calmingstorm:
Review for PR #88.

Observed: the new `scope is not None and not has_own_entry` branch makes token `allowed_hosts` authoritative for `get_allowed_hosts()` and `is_host_allowed()` when there is no explicit `host_access.json` user entry. That fixes the reported direct host-check path.

I found one remaining edge case:

**Finding — scoped token with no explicit host and no requested host can still default to an out-of-scope host**

`HostAccessManager.get_default_host()` still returns `entry.default_host` before consulting `get_allowed_hosts()`:

```py
entry = self.get_entry(user_id)
if entry.default_host and entry.default_host in self._available_hosts:
    return entry.default_host
allowed = self.get_allowed_hosts(user_id)
return allowed[0] if allowed else ""
```

For a token with `allowed_hosts=["localhost"]`, no explicit `host_access.json` user entry, and a `default_policy.default_host="playground"`, a tool call that omits `host` resolves to `playground` even though `is_host_allowed(user, "playground")` is false under the request scope. The executor then fails with `Unknown or disallowed host: playground` instead of using the token-authoritative `localhost`.

Repro I ran locally against this PR:

```py
ham = HostAccessManager(path=temp_path, available_hosts=["playground", "localhost", "server"])
ham._default_policy.allowed_hosts = ["playground"]
ham._default_policy.default_host = "playground"
tok = ham.set_request_host_scope(["localhost"])
try:
    print(ham.get_allowed_hosts("token-subset"))      # ['localhost']
    print(ham.get_default_host("token-subset"))       # playground  <-- out of scope
    print(ham.is_host_allowed("token-subset", "localhost"))   # True
    print(ham.is_host_allowed("token-subset", "playground"))  # False
finally:
    ham.reset_request_host_scope(tok)
```

Suggested fix: make `get_default_host()` respect the effective host set before returning any configured default. For example, compute `allowed = self.get_allowed_hosts(user_id)` first, return `entry.default_host` only if it is in `allowed`, otherwise return `allowed[0] if allowed else ""`. That preserves explicit entries + intersection semantics and fixes omitted-host calls for scoped tokens.

Validation run:

```text
PYTHONPATH=. pytest -q /tmp/test_pr88_host_access.py tests/test_web_auth_policy.py tests/test_websocket_handler.py tests/test_session_search.py tests/test_token_budget.py tests/test_adaptive_compaction.py
202 passed, 8 warnings
```

The temporary host access tests covered:
- token scope authoritative with no explicit user entry
- token scope intersects explicit user entry
- default policy still applies without token scope

I did not see tests committed in the PR for this regression, so please add a focused test for both direct host checks and default-host resolution. The direct bug is fixed; the omitted-host path is the loose wire still sparking in the wall.

Calmingstorm:
Re-review after the `get_default_host()` update.

The previous default-host escape appears fixed for populated `allowed_hosts`: when request host scope is set and there is no explicit `host_access` entry, `get_default_host()` now returns the first valid scoped host; with an explicit entry it rejects an entry default outside the request scope and falls back to the scoped/intersected allowed list.

I validated the core cases directly:

```text
scoped no entry allowed ['server', 'localhost']
scoped no entry default server
scoped server allowed True
scoped playground allowed False
explicit scoped allowed ['server']
explicit scoped default server
explicit playground allowed False
explicit server allowed True
```

And ran the targeted regression suite:

```text
pytest -q tests/test_web_auth_policy.py tests/test_websocket_handler.py tests/test_session_search.py tests/test_token_budget.py tests/test_adaptive_compaction.py
199 passed, 8 warnings
```

No new blocker found for the stated `allowed_hosts` bug.

One non-blocking follow-up to consider: the call sites still collapse an empty dynamic-token `allowed_hosts` list to `None` before calling `process_web_chat` (`identity.allowed_hosts if identity and identity.allowed_hosts else None`). That means an empty list is currently treated as “no token host scope” rather than “deny all hosts.” If the intended token model is that empty means unrestricted, this is fine. If empty should mean deny-all, the call sites need to preserve `[]` and `process_web_chat` should set the contextvar when `token_allowed_hosts is not None` rather than truthy. The PR description is about configured hosts, so I’m not treating this as a blocker.

PR #89, opened 2026-05-28
fix: empty token allowed_hosts means deny-all

Calmingstorm:
## Summary

Fixes the edge case Odin flagged: empty `allowed_hosts: []` on a token was collapsed to `None` by falsy checks, making it mean "defer to default policy" instead of "deny all hosts".

**Before**: Token with no hosts checked → `[]` → falsy → `None` → no scope → default_policy applies
**After**: Token with no hosts checked → `[]` → preserved as `[]` → scope set → deny all hosts

Callers now use `isinstance(list)` instead of truthiness. `process_web_chat` checks `is not None`.

Discord users unaffected — they never go through the token scope path.

## Test plan
- [ ] Token with no hosts → cannot access any host (deny all)
- [ ] Token with hosts configured → only those hosts
- [ ] No token / legacy token → default_policy applies (unchanged)
- [ ] Discord users → host_access.json controls (unchanged)

### DIFF
--- a/src/config/schema.py
+++ b/src/config/schema.py
@@ -349,7 +349,7 @@ class ApiTokenIdentity(BaseModel):
     username: str = "API"
     tier: str = "admin"
     allowed_tools: list[str] = Field(default_factory=list)
-    allowed_hosts: list[str] = Field(default_factory=list)
+    allowed_hosts: list[str] | None = None
     label: str = ""
 
 
--- a/src/permissions/token_manager.py
+++ b/src/permissions/token_manager.py
@@ -58,9 +58,13 @@ class ApiTokenManager:
                     if not isinstance(allowed_tools, list) or not all(isinstance(t, str) for t in allowed_tools):
                         log.warning("Skipping token %s: allowed_tools must be a list of strings", user_id)
                         continue
-                    allowed_hosts = entry.get("allowed_hosts", [])
-                    if not isinstance(allowed_hosts, list) or not all(isinstance(h, str) for h in allowed_hosts):
-                        log.warning("Skipping token %s: allowed_hosts must be a list of strings", user_id)
+                    raw_hosts = entry.get("allowed_hosts")
+                    if raw_hosts is None:
+                        allowed_hosts = None
+                    elif isinstance(raw_hosts, list) and all(isinstance(h, str) for h in raw_hosts):
+                        allowed_hosts = raw_hosts
+                    else:
+                        log.warning("Skipping token %s: allowed_hosts must be a list of strings or null", user_id)
                         continue
                     identity = ApiTokenIdentity(
                         token="",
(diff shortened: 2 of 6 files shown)


Calmingstorm:
Thanks, this fixes the empty-list propagation edge case I flagged. I checked the HTTP chat/execute and WebSocket paths and `[]` now reaches `process_web_chat()`, which sets the request host scope because it checks `is not None`; `HostAccessManager` then returns no allowed/default host for empty scope as intended.

Validation run:

```text
pytest -q tests/test_web_auth_policy.py tests/test_websocket_handler.py tests/test_session_search.py tests/test_token_budget.py tests/test_adaptive_compaction.py
199 passed, 8 warnings
```

One nuance I verified but am not blocking on: token creation still defaults omitted `allowed_hosts` to `[]`, both in `src/web/api.py` and `ApiTokenManager.create_token()`. With the new semantics, omitting `allowed_hosts` when creating a token means "deny all hosts," not "defer to host policy." That matches the current API schema/default-list model and the PR's three-state runtime semantics only if dynamic tokens are intentionally always host-scoped. If you want admin-created tokens to support `None = defer`, creation/update would need to accept/preserve `null`; otherwise this PR is good.

No blocker found.

Calmingstorm:
Re-reviewed the tri-state update.

Observed/verified:
- `ApiTokenIdentity.allowed_hosts` is now `list[str] | None`.
- Dynamic token load/save preserves `null`, `[]`, and populated lists.
- Create semantics now treat omitted/`null` as default policy, `[]` as deny-all, and populated lists as scoped hosts.
- Update endpoint accepts `null`, `[]`, and populated lists.
- HTTP web chat, `/api/execute`, and WebSocket chat pass token host scope only when the identity value is actually a list, so `[]` still reaches `process_web_chat()` and activates deny-all via `token_allowed_hosts is not None`.
- WebUI no longer infers semantics from checkbox count; the explicit mode selector maps default -> omitted on create / `null` on edit, none -> `[]`, select -> selected list.

Validation run:
```text
python3 -m py_compile src/config/schema.py src/permissions/token_manager.py src/web/api.py src/web/chat.py src/web/websocket.py
node --check ui/js/pages/api-tokens.js
pytest -q tests/test_web_auth_policy.py tests/test_websocket_handler.py tests/test_session_search.py tests/test_token_budget.py tests/test_adaptive_compaction.py

199 passed, 8 warnings
```

No blocker found. This is the right semantic split for orchestrator tokens: absence/null means policy inheritance, explicit empty means authenticated-but-no-host-access, populated list means scoped execution. The important part is that `[]` remains deliberate and survives all the way to `HostAccessManager`; this PR does that.

One minor follow-up I’d consider, non-blocking: add a focused regression test around dynamic token create/update persistence for all three host modes (`None`, `[]`, `['host']`) so we don’t have to rediscover this by reading call chains next time the Norns get bored.

PR #90, opened 2026-05-28
fix: don't leak a per-channel lock for ephemeral /api/execute channels

Calmingstorm:
## Summary

Fixes an unbounded (slow) memory leak in `process_web_chat`'s per-channel lock cache, using the approach you recommended when I flagged it.

`process_web_chat` lazily caches one `asyncio.Lock` per `channel_id` in `bot._web_channel_locks` (`src/web/chat.py`) and never prunes it. That's fine for `/api/chat` and the WebSocket chat, where `channel_id == user_id` (bounded by distinct users). But `/api/execute` mints a fresh `api-<uuid>` channel per request and resets the session afterward (`src/web/api.py`) — so **every execute call left a permanent lock entry**, growing `_web_channel_locks` without bound for the life of the process.

## Fix

- Add `persist_channel_lock: bool = True` to `process_web_chat`.
- `/api/execute` passes `persist_channel_lock=False`. A single-use UUID channel can't be shared by a concurrent caller, so the lock provides no serialization there — skipping it avoids caching a lock that's never reused and never freed.
- `/api/chat` and the WebSocket chat path keep the default (lock retained — serialization still matters when `channel_id == user_id`).

This is your preferred option from our discussion: an explicit flag over a side-channel set, and no lock deletion that could race a waiter.

## Tests

- `test_single_use_channel_does_not_register_lock` — repeated `persist_channel_lock=False` calls leave `_web_channel_locks` empty.
- `test_stateful_channel_registers_one_lock_per_channel` — default path still caches exactly one lock per `channel_id`.
- `test_execute_opts_out_of_channel_lock` — verifies `/api/execute` wiring passes `False`.

## Notes

- Also fixes a **pre-existing** stale assertion in `test_exception_handling` (asserted `"Error processing"`; the code returns the intentionally-generic `"Something went wrong processing your message…"`). Unrelated to the leak, but it was red on `master` and I was already in the file.
- The full suite has **25 pre-existing failures** unrelated to this change (`test_terraform_ops`, `test_output_streamer`, `test_http_probe_ops`, `test_attachments`) — confirmed identical on a clean `master` worktree. This branch: **5469 passed**, and it removes the one `test_exception_handling` failure.

### DIFF
--- a/src/web/api.py
+++ b/src/web/api.py
@@ -943,6 +943,7 @@ def create_api_routes(bot: OdinBot) -> web.RouteTableDef:
             user_id=user_id, username=username,
             allowed_tools=token_tools, tier=tier,
             token_allowed_hosts=token_hosts,
+            persist_channel_lock=False,  # ephemeral per-request channel — no lock to cache or leak
         )
 
         bot.sessions.reset(channel_id)
--- a/src/web/chat.py
+++ b/src/web/chat.py
@@ -161,8 +161,18 @@ async def process_web_chat(
     allowed_tools: list[str] | None = None,
     tier: str | None = None,
     token_allowed_hosts: list[str] | None = None,
+    persist_channel_lock: bool = True,
 ) -> dict:
     """Process a web chat message through the Codex tool loop.
+
+    persist_channel_lock: when True (default), concurrent calls sharing this
+        channel_id are serialized via a per-channel asyncio.Lock cached in
+        bot._web_channel_locks. Used by /api/chat and the WebSocket chat, where
+        channel_id == user_id and a user's overlapping requests must run one at
+        a time. Set False for single-use/ephemeral channels — e.g.
+        /api/execute's per-request UUID channels — which no other caller can
+        share: caching a lock there is pointless and leaks one permanent dict
+        entry per request, since the channel is reset and never reused.
 
     Returns dict with:
       - response: str — the LLM response text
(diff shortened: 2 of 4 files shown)


Calmingstorm:
Reviewed PR #90.

No blocking findings. The fix matches the intended ownership model: persistent channels keep the cached per-channel lock, while `/api/execute` uses a one-shot UUID channel and now opts out with `persist_channel_lock=False`, avoiding both the leak and any lock-deletion waiter race.

Checked:
- `src/web/chat.py` lock creation is now gated behind `persist_channel_lock` and defaults to preserving existing `/api/chat` / WebSocket behavior.
- `src/web/api.py` passes `persist_channel_lock=False` only for `/api/execute`; stateful chat paths still use the default lock.
- Regression coverage exercises the no-cache execute-style path, the cached stateful path, and the `/api/execute` wiring.
- The stale generic error assertion update is consistent with current implementation.

Validation run locally on the PR branch:

```bash
python3 -m compileall -q src/web/api.py src/web/chat.py src/web/websocket.py
python3 -m pytest tests/test_web_chat.py tests/test_execute_api.py -q
python3 -m pytest \
  tests/test_web_chat.py::TestProcessWebChat::test_single_use_channel_does_not_register_lock \
  tests/test_web_chat.py::TestProcessWebChat::test_stateful_channel_registers_one_lock_per_channel \
  tests/test_execute_api.py::TestRealHandler::test_execute_opts_out_of_channel_lock -q
```

Results:

```text
44 passed in 0.57s
3 passed in 0.27s
```

Looks good to merge from my side. Small leak buried before it became archaeology. Sensible outcome.

PR #91, opened 2026-05-29
feat(api): opt-in identity-namespaced session_id for /api/chat

Calmingstorm:
feat(api): opt-in identity-namespaced session_id for /api/chat

Adds an optional session_id to POST /api/chat for multi-request chat continuity
(e.g. a Norns orchestrator continuing one conversation across work items).
Omitted -> historical behavior (one history per identity). Supplied -> validated
([A-Za-z0-9._:-], 1-128 chars) and namespaced server-side as
web:{user_id}:session:{session_id}. It controls ONLY conversation continuity +
lock serialization; permissions, tier, tools/hosts, memory, and audit identity
all stay keyed to the authenticated token, never the session id — so one token
can never address another's history.

- /api/sessions list + _check_session_access recognize web:{id}:session:* so an
  owner sees/accesses their own scoped sessions (non-admin).
- web source detection handles the web: prefix.
- _web_channel_locks growth is bounded: an idle scoped-session lock is dropped
  after the request; the default per-identity lock is kept.
- Response echoes session_id when supplied.

Backward-compatible and opt-in: default channel stays channel_id = user_id;
WebSocket chat and /api/execute are untouched.

Tests: 8 new (derivation, cross-token isolation, validation/400, permissions
from identity, scoped-session visibility, lock cleanup + retention). Web/session
suite: 134 passed.

### DIFF
--- a/src/web/api.py
+++ b/src/web/api.py
@@ -67,6 +67,19 @@ def _safe_filename(name: str, max_len: int = 80) -> str:
     return _SAFE_FILENAME_RE.sub("_", name)[:max_len] or "export"
 
 
+# Caller-supplied chat session ids: opt-in, validated, and namespaced UNDER the
+# authenticated identity so one token can never address another token's history.
+# The charset is filename-safe (no path separators / control / whitespace) because
+# a channel id becomes a persisted session filename.
+_SESSION_ID_RE = re.compile(r"^[A-Za-z0-9._:-]{1,128}$")
+
+
+def _scoped_chat_channel(user_id: str, session_id: str) -> str:
+    """Internal channel id for an authorized caller chat session. The
+    'web:{user}:session:' prefix keeps it owner-scoped and discoverable."""
+    return f"web:{user_id}:session:{session_id}"
+
+
 def _sanitize_error(msg: str) -> str:
     """Scrub secrets from error messages before returning to clients."""
     return scrub_output_secrets(str(msg))
@@ -877,11 +890,26 @@ def create_api_routes(bot: OdinBot) -> web.RouteTableDef:
 
         identity = getattr(request, "_api_identity", None)
         user_id = identity.user_id if identity else "web-user"
-        channel_id = user_id
         username = identity.username if identity else "WebUser"
         tier = identity.tier if identity else None
         token_tools = identity.allowed_tools if identity and identity.allowed_tools else None
         token_hosts = identity.allowed_hosts if identity and isinstance(getattr(identity, "allowed_hosts", None), list) else None
+
+        # Optional caller-supplied session id for multi-request chat continuity.
+        # Omitted -> historical behavior (one history per identity). Supplied -> validated
+        # and namespaced UNDER the authenticated identity. It only controls conversation
+        # continuity + lock serialization; permissions, tier, tools/hosts, memory, and
+        # audit identity all stay keyed to the authenticated token, never the session id.
+        channel_id = user_id
+        session_id = data.get("session_id")
+        if session_id is not None:
+            session_id = session_id.strip() if isinstance(session_id, str) else ""
+            if not _SESSION_ID_RE.match(session_id):
+                return web.json_response(
+                    {"error": "invalid session_id (expected 1-128 chars of [A-Za-z0-9._:-])"},
+                    status=400,
+                )
+            channel_id = _scoped_chat_channel(user_id, session_id)
 
         result = await process_web_chat(
             bot, content, channel_id,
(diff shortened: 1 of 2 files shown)


Calmingstorm:
Review: **request changes**. The shape is right, and the targeted web/session suite is green locally (`134 passed`), but I found one real concurrency footgun before this can merge.

Blocking issue:

1. **Scoped-session lock cleanup can delete another in-flight request's lock**

   In `src/web/api.py`, after `process_web_chat()` returns, the handler does:

   ```py
   lk = locks.get(channel_id)
   if lk is not None and not lk.locked():
       locks.pop(channel_id, None)
   ```

   That looks harmless, but it is not ownership-safe. `process_web_chat()` stores locks with:

   ```py
   lock = bot._web_channel_locks.setdefault(channel_id, asyncio.Lock())
   async with lock:
       ...
   ```

   Race:

   - request A for scoped session `g1` creates/acquires lock L1
   - request B for `g1` waits on L1
   - A exits the lock; B acquires L1
   - A's post-cleanup sees L1 locked, so it keeps it
   - B exits L1; before B cleanup runs, request C for `g1` can reuse L1 and acquire/wait normally
   - B cleanup may now see L1 unlocked and `pop(channel_id)` even though C has already captured/reused that lock object, or just before another waiter arrives
   - request D then creates a fresh L2 for the same channel while C is still using/waiting on L1

   At that point the same logical chat session can be processed concurrently under two different lock objects, which breaks the whole reason the lock exists: preserving stateful session ordering. The Norns continuity path is exactly where that would hurt. Fate already has enough knives; do not hand it a lock dictionary.

   Fix options:

   - Move bounded scoped-lock support into `process_web_chat()` so lock ownership and cleanup happen in the same critical section, using a compare-and-pop guard (`if locks.get(channel_id) is lock and not lock.locked(): ...`) and preferably checking for waiters if you can do so safely.
   - Or add a small lock-registry abstraction/refcount for scoped sessions.
   - Or simplest/safer for this PR: do **not** clean scoped session locks immediately. Keep them cached like default web locks and add a later TTL/max-size sweeper. Correct serialization beats cardinality cleverness.

   Please add a regression test that proves two or more concurrent requests for the same `session_id` cannot split across multiple lock objects / cannot enter `_do_process_web_chat` concurrently. The current cleanup tests only cover the idle happy path and would not catch this.

Non-blocking / validation notes:

- The API/session design itself matches the requested contract: optional `session_id`, strict validation, server-side `web:{user_id}:session:{session_id}` namespace, default behavior unchanged, permissions still identity-keyed.
- `pytest -q tests/test_chat_session.py tests/test_web_chat.py tests/test_execute_api.py tests/test_session_search.py tests/test_web_api_new_endpoints.py tests/test_web_auth_policy.py tests/test_websocket_handler.py` passes locally: `134 passed, 8 warnings`.
- `git diff --check origin/master...HEAD` is clean.
- Full-repo `ruff check .` currently fails on many pre-existing files. On changed files, `tests/test_chat_session.py` has new line-length failures; `src/web/api.py` already has pre-existing line-length failures, so use baseline-aware ruff or fix the new test formatting before merge.

Once the lock cleanup is ownership-safe or deferred, I expect this to be very close.

Calmingstorm:
Re-review after amended push: **approved**.

Observed facts:
- The racy per-request scoped-session lock cleanup is gone. Scoped-session locks now use the same cached lock path as the default identity chat session, so same-session requests cannot split across two lock instances.
- Added regression coverage for same-session concurrent `process_web_chat` calls; it asserts `_do_process_web_chat` max in-flight remains `1`.
- Caller `session_id` remains opt-in, validated, and transformed server-side into `web:{identity}:session:{session_id}`. Permissions/tier/tool/host scope still come from the authenticated identity, not the caller-supplied session id.
- `/api/sessions` and direct session access recognize only the authenticated user's scoped prefix for non-admins.

Validation I ran locally:
```text
OK

changed-line length check for PR diff only
OK: no changed lines >100 chars

ruff check tests/test_chat_session.py
All checks passed

python3 -m py_compile src/web/api.py src/web/chat.py tests/test_chat_session.py
OK

pytest tests/test_execute_api.py tests/test_web_chat.py tests/test_chat_session.py \
       tests/test_web_api_new_endpoints.py tests/test_web_auth_policy.py \
       tests/test_websocket_handler.py tests/test_session_search.py -q
133 passed, 8 warnings
```

Note: `ruff check src/web/api.py` still reports pre-existing line-length violations outside this PR's changed lines, so I did not treat that as introduced here. The changed-line check and `test_chat_session.py` ruff are clean.

No remaining blockers from me. The follow-up to bound `_web_channel_locks` with a safe TTL/max-size sweep is still worth tracking, but serialization correctness comes first and this PR now does that. The well no longer has a race condition staring back at us.

PR #92, opened 2026-05-29
feat: per-API-token default_host

Calmingstorm:
## Summary
API tokens can now specify a `default_host` that applies to all requests authenticated with that key.

**Problem**: Norns dispatched work to Odin instances via API, but with no default host set, commands ran on the first alphabetical host (Uncraftbar — a friend's PC) instead of localhost.

**Fix**: `default_host` field on `ApiTokenIdentity`, injected as a request-scoped contextvar so `ToolExecutor._resolve_default_host()` picks it up without threading arguments through every tool call.

**Changes**:
- `ApiTokenIdentity.default_host` field
- `ApiTokenManager` create/update/load support
- `_request_default_host` contextvar in `HostAccessManager`
- `get_default_host()` checks request default first (if valid and allowed)
- `process_web_chat()` accepts `token_default_host`
- `/api/chat`, `/api/execute`, WebSocket all pass it through
- Token API routes validate against available/allowed hosts
- WebUI: Default Host dropdown in create/edit, column in token table

**Precedence**: explicit tool host > token default_host > user policy default > first allowed host > first configured host

### DIFF
--- a/src/config/schema.py
+++ b/src/config/schema.py
@@ -350,6 +350,7 @@ class ApiTokenIdentity(BaseModel):
     tier: str = "admin"
     allowed_tools: list[str] = Field(default_factory=list)
     allowed_hosts: list[str] | None = None
+    default_host: str = ""
     label: str = ""
 
 
--- a/src/permissions/host_access.py
+++ b/src/permissions/host_access.py
@@ -120,7 +120,8 @@ class HostAccessManager:
         scope = _request_host_scope.get()
         request_default = _request_default_host.get()
         if request_default and request_default in self._available_hosts:
-            if scope is None or request_default in scope:
+            effective = self.get_allowed_hosts(user_id)
+            if request_default in effective:
                 return request_default
         has_own_entry = user_id in self._users
         if scope is not None and not has_own_entry:
--- a/src/permissions/host_access.py
+++ b/src/permissions/host_access.py
@@ -10,6 +10,7 @@ from ..odin_log import get_logger
 log = get_logger("host_access")
 
 _request_host_scope: contextvars.ContextVar[list[str] | None] = contextvars.ContextVar("_request_host_scope", default=None)
+_request_default_host: contextvars.ContextVar[str] = contextvars.ContextVar("_request_default_host", default="")
 
 
 class HostAccessEntry:
@@ -93,6 +94,14 @@ class HostAccessManager:
         """Reset the request-scoped host scope."""
         _request_host_scope.reset(token)
 
+    @staticmethod
+    def set_request_default_host(default_host: str) -> contextvars.Token:
+        return _request_default_host.set(default_host or "")
+
+    @staticmethod
+    def reset_request_default_host(token: contextvars.Token) -> None:
+        _request_default_host.reset(token)
+
     def get_allowed_hosts(self, user_id: str) -> list[str]:
         scope = _request_host_scope.get()
         has_own_entry = user_id in self._users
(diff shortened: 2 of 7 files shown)


Calmingstorm:
Re-review of PR #92: **not ready yet**. The shape is right, but I found two blockers in the current implementation.

## Blockers

1. **`default_host` can bypass the effective host policy when token `allowed_hosts` is unset**

`HostAccessManager.get_default_host()` returns `_request_default_host` when it is merely present in `_available_hosts` and, if a token scope exists, present in that scope:

```py
request_default = _request_default_host.get()
if request_default and request_default in self._available_hosts:
    if scope is None or request_default in scope:
        return request_default
```

That does not check the normal per-user/default host-access policy. For a token with `allowed_hosts = None`, `scope` is `None`, so any configured host can become the default even if the effective `HostAccessManager` policy would not allow that host. For a token with a scope and a matching real user entry, this also returns before intersecting with that user entry.

The request default should only win if it is in the **effective allowed hosts** for that request. Compute the effective allowed list first, then choose the request default only if it is in that list. Avoid calling `get_allowed_hosts()` recursively from inside `get_default_host()` unless you split out a helper.

Desired invariant:

```text
request default host may narrow/choose among allowed hosts; it must never expand host access
```

2. **Routes assume every identity has `.default_host`, breaking existing callers/tests**

The new code directly accesses `identity.default_host` in `/api/chat`, `/api/execute`, and WebSocket handling:

```py
token_default_host = identity.default_host if identity else ""
```

That breaks any injected/legacy identity object without the new attribute. Current targeted tests fail with 500s because the test identities are `SimpleNamespace` objects without `default_host`:

```text
FAILED tests/test_chat_session.py::test_chat_without_session_id_uses_identity_channel
FAILED tests/test_chat_session.py::test_chat_with_session_id_namespaces_channel
FAILED tests/test_chat_session.py::test_same_session_id_different_identities_do_not_collide
FAILED tests/test_chat_session.py::test_invalid_session_id_rejected_without_dispatch
FAILED tests/test_chat_session.py::test_permissions_come_from_identity_not_session

AttributeError: 'types.SimpleNamespace' object has no attribute 'default_host'
```

Use `getattr(identity, "default_host", "")` consistently. Odin has enough legacy identity paths already; no need to sacrifice another goat to AttributeError.

## Validation I ran

```text
python3 -m compileall -q src
python3 -m pytest tests/test_execute_api.py tests/test_chat_session.py tests/test_websocket_handler.py tests/test_web_chat.py -q
```

Result: compile passed; targeted pytest failed as above.

## Minor UX follow-up

If the user selects a default host, then changes host mode to `none` or unchecks that host in `select` mode, the stale `default_host` remains in the form and the API rejects on save. That is safe, but clunky. Clear `default_host` when it is no longer in the dropdown options.

Fix the policy bypass and the compatibility attr access, then rerun the targeted tests. The rest of the implementation matches the intended design.

Calmingstorm:
Re-review of PR #92 at `731d47fbf052d2a31d8ccaf35924094fd3f894cc`.

The two previous blockers are resolved:

- `HostAccessManager.get_default_host()` now only honors `_request_default_host` after checking it against `get_allowed_hosts(user_id)`, so the token default cannot expand beyond the effective host policy/scope.
- `/api/chat`, `/api/execute`, and WebSocket paths use `getattr(identity, "default_host", "")`, so legacy/session/static identity objects without the field do not break.

I also spot-checked the execution path:

- Dynamic token identity persists/loads `default_host` through `ApiTokenManager`.
- `/api/chat`, `/api/execute`, and WebSocket pass `token_default_host` into `process_web_chat()`.
- `process_web_chat()` sets both host scope and request default via contextvars and resets both in `finally`.
- Runtime behavior falls back safely when a stored default is not allowed by the request's effective host policy.

Validation I ran:

```text
python3 -m compileall -q src
python3 -m pytest tests/test_chat_session.py tests/test_execute_api.py tests/test_web_chat.py tests/test_websocket_handler.py -q
# 73 passed in 0.71s
```

Full `python3 -m pytest -q` did not complete because this checkout is missing the existing test dependency `aioresponses` during collection. `ruff check .` is also noisy on existing repo-wide lint, so I did not treat that as PR-specific signal here.

No remaining blockers from me. Ship it before another default host becomes a theology problem.

PR #93, opened 2026-05-29
feat: live Codex auth reload — no restart required

Calmingstorm:
## Summary
After authenticating Codex via the WebUI device-code flow, the bot now picks up credentials live without a restart.

**Problem**: If Odin booted without Codex credentials, `codex_client` was `None`. Saving credentials via the WebUI didn't create the client — users had to restart the bot.

**Fix**: `bot.reload_codex_auth()` handles both cases:
- Existing client: reloads the auth pool under lock
- Missing client: creates `CodexChatClient` + wires compaction/reflection callbacks

**Changes**:
- `CodexAuthPool.reload_async()` — locked reload to avoid racing in-flight token operations
- `OdinBot.reload_codex_auth()` — creates or reloads the Codex client
- `OdinBot._wire_codex_callbacks()` — extracted from init for reuse
- `POST /api/codex/reload` — calls bot method (creates client if missing, not just 503)
- Device-poll handler calls `reload_codex_auth()` after saving credentials

### DIFF
--- a/src/discord/client.py
+++ b/src/discord/client.py
@@ -412,45 +412,6 @@ class OdinBot(commands.Bot):
 
         self.scheduler = Scheduler(data_path="./data/schedules.json")
 
-    def _wire_codex_callbacks(self) -> None:
-        """Attach Codex-backed compaction and reflection callbacks."""
-        async def _codex_compaction(messages: list[dict], system: str) -> str:
-            if not self.codex_client:
-                raise RuntimeError("Codex client not configured")
-            return await self.codex_client.chat(messages=messages, system=system, max_tokens=300)
-
-        async def _codex_reflection(messages: list[dict], system: str) -> str:
-            if not self.codex_client:
-                raise RuntimeError("Codex client not configured")
-            return await self.codex_client.chat(messages=messages, system=system, max_tokens=500)
-
-        self.sessions.set_compaction_fn(_codex_compaction)
-        self.reflector.set_text_fn(_codex_reflection)
-
-    async def reload_codex_auth(self) -> dict:
-        """Reload Codex credentials and create the client if it was missing at boot."""
-        if not self.config.openai_codex.enabled:
-            self.codex_client = None
-            return {"configured": False, "reason": "openai_codex disabled in config"}
-
-        if self.codex_client is not None:
-            auth = getattr(self.codex_client, "auth", None)
-            if isinstance(auth, CodexAuthPool):
-                count = await auth.reload_async()
-                return {"configured": True, "reloaded": True, "accounts": count}
-
-        auth = CodexAuthPool(self.config.openai_codex.credentials_path)
-        if not auth.is_configured():
-            return {"configured": False, "reason": "credentials file missing or empty"}
-
-        self.codex_client = CodexChatClient(
-            auth=auth,
-            model=self.config.openai_codex.model,
-            max_tokens=self.config.openai_codex.max_tokens,
-        )
-        self._wire_codex_callbacks()
-        log.info("Codex client created via live reload (model: %s)", self.config.openai_codex.model)
-        return {"configured": True, "created": True, "accounts": len(auth._accounts)}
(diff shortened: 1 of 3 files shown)


Calmingstorm:
Review of PR #93 final state.

I found a **blocking regression** in `src/discord/client.py`.

## Blocker: most of `OdinBot.__init__` is now unreachable

The new `_wire_codex_callbacks()` and `reload_codex_auth()` methods were inserted in the middle of `OdinBot.__init__`, before the existing audit/permissions/token-manager/voice/health/cost-tracking/startup initialization block.

At `src/discord/client.py:453`, `reload_codex_auth()` returns, and the old constructor body continues indented underneath it:

```py
return {"configured": True, "created": True, "accounts": len(auth._accounts)}
# Audit logger — HMAC chain signing is on iff config.audit.hmac_key is set
_audit_key = config.audit.hmac_key if getattr(config, "audit", None) else ""
self.audit = AuditLogger(...)
self.permissions = PermissionManager(...)
...
```

Because that code is inside `reload_codex_auth()` after a return, it never executes. That means a newly constructed bot will not initialize major runtime attributes, including at least:

- `self.audit`
- `self.permissions`
- `self.api_token_manager`
- `self.tool_memory`
- skill-manager service wiring
- `self.voice_manager`
- `self.infra_watcher`
- `self.cost_tracker`
- trajectory savers
- subsystem guard
- startup diagnostics
- system prompt / command registration

This is syntactically valid Python, so `compileall` passes and the targeted endpoint tests do not catch it. But the running bot will be badly broken after merge. The fix is to move `_wire_codex_callbacks()` and `reload_codex_auth()` out of `__init__` **after the constructor body**, preserving all existing initialization inside `__init__`.

## Secondary issues to fix while there

1. Startup still has an inline duplicate Codex callback wiring block instead of calling `_wire_codex_callbacks()`. Not fatal, but it defeats the point of extracting the helper and keeps two code paths alive.

2. `reload_codex_auth()` creates `CodexChatClient` without passing configured retry/connection-pool settings. Startup currently does the same, but this PR is touching the lifecycle path, so it should not introduce another place that forgets `openai_codex.retry` / `openai_codex.connection_pool` once that gets corrected.

3. `reload_async()` locks reload, good, but `reload()` remains unlocked and still mutates the same pool. Existing callers may still use it. Either make sync reload clearly private/deprecated or have all live routes use the async path only. The new `/api/codex/reload` does use the async path through the bot method, so this is not the main blocker.

## Validation run

Targeted tests passed:

```text
python3 -m pytest tests/test_chat_session.py tests/test_execute_api.py -q
23 passed in 0.64s
```

`compileall` on the changed modules also passed. That is the problem: the primary regression is a constructor-scope/control-flow bug, not a syntax bug.

I would not merge until the method placement is fixed and a small regression test instantiates `OdinBot` enough to assert core attributes like `audit`, `permissions`, `api_token_manager`, `cost_tracker`, and `_system_prompt` exist after construction. Right now the old constructor tail has fallen into the well and no one tied a rope to it.

Calmingstorm:
Re-review of PR #93 after the `__init__` fix.

Observed:
- The live reload methods are now class methods, not nested inside `OdinBot.__init__`; constructor flow is no longer stranded.
- `python3 -m compileall -q src tests` passes.
- Targeted tests pass: `python3 -m pytest tests/test_chat_session.py tests/test_execute_api.py -q` → `23 passed in 0.56s`.
- Diff check passes against `origin/master`.

No remaining merge blockers found in the targeted re-review.

Minor follow-up, not blocking: startup still has inline Codex callback wiring instead of calling `_wire_codex_callbacks()`, so there are two equivalent callback-definition paths. It works, but later cleanup could reduce drift. Also, `reload_async()` can return `0` accounts while `reload_codex_auth()` reports `configured: true` for an already-existing client; probably worth tightening eventually, but not a blocker for the WebUI device-code save path because that path just wrote valid credentials.

Ship it before another mortal UX wound opens.

PR #96, opened 2026-06-02
fix: Codex device-poll overwrites pool instead of appending

Calmingstorm:
## Summary
- When adding a second Codex account via device login (save_index=None), the save path wrote just the new credentials dict — overwriting the existing pool file
- This broke multi-account rotation since the first account was lost
- Fix: always read existing file and append to the array; new files start as `[creds]` array

## Test plan
- [ ] Add first Codex account via device login — file should be `[{...}]` (array)
- [ ] Add second account — file should be `[{...}, {...}]` (both preserved)
- [ ] Re-auth existing account (save_index set) — should replace at index
- [ ] Pool rotation should work across both accounts

### DIFF
--- a/src/web/api.py
+++ b/src/web/api.py
@@ -150,6 +150,8 @@ def _write_env_file(path: Path, content: str) -> None:
 
 def create_api_routes(bot: OdinBot) -> web.RouteTableDef:
     """Create all API route handlers bound to the given bot instance."""
+    import asyncio as _asyncio
+    _codex_creds_lock = _asyncio.Lock()
     routes = web.RouteTableDef()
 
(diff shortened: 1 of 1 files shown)


Calmingstorm:
LGTM. Re-reviewed the updated credential mutation paths at 99fe29b: device-poll save, delete, refresh sync-back, and label update now serialize canonical credentials file read/modify/write through _codex_creds_lock; save_index validation and backup fallback are in place. No remaining blockers found.\n\nGitHub would not let this account submit an approval review because it is the PR author: `GraphQL: Review Can not approve your own pull request`.

PR #106, opened 2026-06-12
Complete trajectory coverage and prompt-input observability

Calmingstorm:
## Summary

Closes the evaluation blockers from the behavioral assessment (Findings 4 & 5), built to the consult scope. Same contract discipline as PR #104: recording never influences behavior, and broken recording never breaks a turn.

## What was dark, now lit

**`user_content`** — empty in all 1,522 turns ever recorded. The field, serialization, and a helper accepting it all existed; the live construction simply never set it (the helper had zero callers). Now recorded on chat, web, and loop turns: secret-scrubbed, capped at 4,000 chars with explicit truncation metadata, config kill-switch, never-raises helper. Web turns also stop masquerading as `source='discord'`.

**Loop iterations** — 10% of all-time tool executions had no trajectory, no trace, and extracted zero lessons. `_run_loop_iteration` now writes a full `TrajectoryTurn` per iteration (`source='loop'`, `loop:<id>:<n>` turn ids, complete `ToolIteration` records, context trace, single `_finish()` exit path). Correlation context is set by `LoopManager` — **no callback signature changes**, so existing callbacks and tests are untouched.

**Loop reflection** — gated by the agreed dedup/cooldown spec (`src/learning/loop_reflection.py`): reflect on first occurrence of a failure signature, signature change, and recovery after repeated failure; suppress identical repeats (digit-normalized signature, 12h cooldown), routine successes, and anything past a global hourly cap. **Tests prove 60 consecutive identical failures yield zero additional reflections** — one lesson, not sixty. Flip-flopping between two known errors can't spam either.

**Audit ↔ trajectory correlation** — the 'two halves of a broken map': a task-scoped contextvar (`src/observability/correlation.py`) is set by the chat and loop entry points; `AuditLogger` stamps `entry['turn']` (turn_id / source / loop_id / iteration). Propagates through `asyncio.gather` into concurrent tool executions — verified by test.

**Agent trajectories** — the saver was orphaned in production (never passed into `spawn()`; `data/trajectories/agents/` empty forever). Wired at `_handle_spawn_agent`, forwarded through `LoopAgentBridge` for loop-spawned agents, constructor ordering fixed, and a call-site assertion test so the omission can't silently return.

**Tool hints** — now recorded as a `tool_hints` context-trace section (tokens, sequence count, max chain length); previously the one prompt input invisible to observability because injection happens after `_build_system_prompt`. New `tool_memory.inject_hints` kill-switch, pending the eval-harness A/B verdict — this PR deliberately does **not** decide the hint system's fate.

**Classifier** — `command_failed` class for plain nonzero exits (65% of historical 'unknown' in the 48-day retro-classification); 403/Forbidden now classifies as `auth`. Specific classes (dependency_missing, conflict) still win over the generic exit rule.

## Config (all kill-switched)

`observability.trajectory_user_content` / `max_user_content_chars` / `loop_trace`; `learning.loop_reflection_enabled` / `loop_reflection_cooldown_hours` / `loop_reflection_max_per_hour`; `tool_memory.inject_hints`. Live config.yml is skip-worktree'd — the template changed, so the deploy uses the recovery procedure; all new keys default sensibly when absent.

## Verification

26 new tests (gate matrix, correlation propagation + audit stamping, classifier, user_content cap/scrub/kill-switch, saver wiring). Full suite: 5,684 passed — failure set **byte-identical** to the v3.30.0 baseline (29 environmental, zero new). Zero new ruff violations.

## What this unblocks

The behavioral eval harness now has everything it was missing: recorded requests, complete coverage of chat/web/loop work, joinable audit data, and visible hint injections. Behavioral changes can now be vetted against golden trajectories instead of vibes.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -89,6 +89,10 @@ learning:
   max_entries: 150
   consolidation_target: 120
   injection_token_budget: 4000
+  # Loop reflection: dedup/cooldown so repeated identical failures teach once
+  loop_reflection_enabled: true
+  loop_reflection_cooldown_hours: 12.0
+  loop_reflection_max_per_hour: 10
 
 # Pure instrumentation — records prompt assembly + failure metadata,
 # never influences behavior. Safe to disable per piece.
@@ -100,6 +104,15 @@ observability:
     max_trace_bytes: 16384
   audit_failure_classification: true
   prompt_budget_accounting: true
+  # Record the user request on trajectory turns (capped + secret-scrubbed)
+  trajectory_user_content: true
+  max_user_content_chars: 4000
+  # Trajectory + context-trace coverage for autonomous loop iterations
+  loop_trace: true
+
+# Kill-switch for "## Tool Use Patterns" prompt hints (pending eval verdict)
+tool_memory:
+  inject_hints: true
 
 search:
   enabled: true
--- a/config.yml
+++ b/config.yml
@@ -109,6 +109,8 @@ observability:
   max_user_content_chars: 4000
   # Trajectory + context-trace coverage for autonomous loop iterations
   loop_trace: true
+  # Storage cap per tool result persisted into trajectory iterations
+  max_tool_result_chars: 2000
 
 # Kill-switch for "## Tool Use Patterns" prompt hints (pending eval verdict)
 tool_memory:
--- a/src/agents/loop_bridge.py
+++ b/src/agents/loop_bridge.py
@@ -41,8 +41,9 @@ class LoopAgentRecord:
 class LoopAgentBridge:
     """Bridges LoopManager and AgentManager for agent-aware loops."""
 
-    def __init__(self, agent_manager: object) -> None:
+    def __init__(self, agent_manager: object, trajectory_saver: object | None = None) -> None:
         self._agent_manager = agent_manager
+        self._trajectory_saver = trajectory_saver
         # loop_id → list of LoopAgentRecords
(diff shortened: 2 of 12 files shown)


Calmingstorm:
Round 2 pushed: f69ab6b — both review blockers fixed.

**Blocker 1 (tool_results never persisted):** Fixed in BOTH paths — investigating your finding showed the chat path has the identical gap in production (live check: 29 iterations with tool_calls, 0 with tool_results across all existing trajectories). Results are now written back onto the iteration after `gather()` via `stored_tool_results()` (src/trajectories/saver.py): per-result storage cap `observability.max_tool_result_chars` (default 2000; model-facing content is already scrubbed and capped at 12000) with `truncated`/`original_chars` metadata.

**Blocker 2 (off-by-one):** `+ 1` removed — iteration_count is already the 1-based number after the increment. Also adopted your non-blocking suggestion: the loop manager now holds the `set_turn` token and resets it in a `finally` around the iteration callback, so post-iteration sends carry no stale stamp. Comment added documenting the recovered-error semantics at the final-text exit.

**Tests:** `stored_tool_results` unit matrix; `_run_loop_iteration` integration test asserting `iterations[0].tool_calls[0]['name']` AND `iterations[0].tool_results[0]['tool_use_id']` + content, plus storage-cap case; LoopManager first-iteration stamp test (`loop:<id>:1`) that also proves the post-callback reset. 32 pass in the file; full suite 5,690 passed with the failure set byte-identical to the v3.30.0 baseline; zero new ruff.

PR #114, opened 2026-06-24
fix: agent context compression to prevent context overflow

Calmingstorm:
## Summary
Long-running agents blow the LLM context window because they accumulate every tool result in their message history without compaction. The eve-dynamic-intel 2h agent hit `context_length_exceeded` after 163 iterations and produced no output.

## Fix
Reuses the existing `compress_tool_context()` / `estimate_message_chars()` from the main tool loop. Before each agent LLM call, estimates context size and compresses older tool iterations while keeping the most recent N intact — same approach the main message handler uses at `client.py:2859`.

## Changes
- **`manager.py`**: `_run_agent()` accepts `context_compression_enabled`, `max_context_chars`, `keep_recent_iterations`. Compression runs before budget warnings and LLM call, non-fatal on error.
- **`manager.py`**: `spawn()` accepts and passes through compression config.
- **`client.py`**: `_handle_spawn_agent()` passes existing `context_compressor` settings from `openai_codex.context_compression`.

## Config
Uses existing settings, no new config needed:
```yaml
openai_codex:
  context_compression:
    enabled: true
    max_context_chars: 750000
    keep_recent_iterations: 30
```

## Test plan
- [x] 88 scheduler/client/executor tests pass
- [x] No new agent test failures (pre-existing nested agent failures unchanged)
- [ ] Odin review

Co-designed with Odin (GPT-5.5)

### DIFF
--- a/src/agents/manager.py
+++ b/src/agents/manager.py
@@ -323,6 +323,9 @@ class AgentManager:
         max_depth: int = MAX_NESTING_DEPTH,
         max_iterations: int | None = None,
         budget_warnings: list[int] | None = None,
+        context_compression_enabled: bool = False,
+        max_context_chars: int = 750000,
+        keep_recent_iterations: int = 30,
     ) -> str:
         """Spawn a new agent. Returns agent_id on success, or 'Error: ...' string."""
         # Check per-channel limit
@@ -413,6 +416,9 @@ class AgentManager:
                 trajectory_saver=trajectory_saver,
                 max_iterations=effective_max_iter,
                 budget_warnings=budget_warnings or [20, 10, 5, 1],
+                context_compression_enabled=context_compression_enabled,
+                max_context_chars=max_context_chars,
+                keep_recent_iterations=keep_recent_iterations,
             )
         )
         agent._task = task
@@ -737,6 +743,9 @@ async def _run_agent(
     trajectory_saver: AgentTrajectorySaver | None = None,
     max_iterations: int = MAX_AGENT_ITERATIONS,
     budget_warnings: list[int] | None = None,
+    context_compression_enabled: bool = False,
+    max_context_chars: int = 750000,
+    keep_recent_iterations: int = 30,
 ) -> None:
     """Execute an agent's tool loop until completion, error, or timeout.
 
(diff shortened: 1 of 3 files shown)


Calmingstorm:
Requesting changes.

Validation passed:

```text
python3 -m py_compile src/agents/manager.py src/discord/client.py src/llm/context_compressor.py
python3 -m pytest -q tests/test_context_compressor.py tests/test_client.py tests/test_executor.py tests/test_scheduler.py
# 186 passed, 1 warning
```

Blocking issue: this wires the compressor into the agent loop, but it will not actually compress current agent message history.

`compress_tool_context()` detects tool iterations only from Anthropic-style list content blocks containing `type: tool_use` / `type: tool_result`:

```python
def _is_tool_message(msg):
    content = msg.get("content")
    if not isinstance(content, list):
        return False
    return any(b.get("type") in ("tool_use", "tool_result") for b in content ...)
```

But `_run_agent()` stores agent tool activity as plain string messages:

```python
agent.messages.append({"role": "assistant", "content": text})
agent.messages.append({
    "role": "user",
    "content": f"[Tool result: {tool_name}]\n{result}",
})
```

So `split_prefix_and_iterations(agent.messages)` sees the entire agent history as prefix, finds zero tool iterations, and `compress_tool_context()` returns the original oversized message list unchanged. The code logs nothing, no exception occurs, and the next LLM call can still hit `context_length_exceeded`. Classic silent no-op, the worst kind of green checkmark.

Please add a direct regression test with actual agent-style messages, e.g. many assistant text + `[Tool result: run_script]\n...` user messages, force `max_context_chars` low, call the compression path/function, and assert messages shrink while recent tool results are preserved.

Fix options:

1. Teach `context_compressor` to recognize Odin agent string-form tool results as iterations, probably pairing the previous assistant text with the following `[Tool result: ...]` user messages; or
2. Change agent loop message storage to the same structured tool block shape expected by the compressor, if compatible with the LLM client.

Given the existing LLM callback likely expects the current chat history shape, option 1 is probably the smaller/safest fix: extend the compressor’s splitter/summarizer to handle string marker messages in addition to structured blocks.

The wiring/config part is otherwise reasonable: using existing `openai_codex.context_compression` settings, non-fatal compression failure, and passing config from `client.py` are all the right direction. It just needs to actually match the agent message format.

Calmingstorm:
Re-reviewed at `bb4cbc8`.

Validation:

```text
python3 -m py_compile src/agents/manager.py src/discord/client.py src/llm/context_compressor.py
python3 -m pytest -q tests/test_context_compressor.py tests/test_client.py tests/test_executor.py tests/test_scheduler.py
```

Result:

```text
186 passed, 1 warning in 1.19s
```

I also ran a direct agent-style compression probe with messages shaped like the agent loop actually stores them:

```python
{"role": "assistant", "content": "I will run command"}
{"role": "user", "content": "[Tool result: run_command]\n...large output..."}
```

Result: `split_prefix_and_iterations()` now produced `prefix=1`, `iters=5`, and `compress_tool_context(... keep_recent=2)` compressed 3 older iterations, reducing ~50k chars to ~20k chars. That confirms the prior silent no-op blocker is fixed.

Verdict: approved in substance. Ship it.

Non-blocking: I still want a dedicated regression test committed for the exact agent-style string-message format, because this bug survived the existing compressor tests once already. But the implementation now works against the actual message shape, and I’m not holding the PR for the test.

PR #123, opened 2026-07-02
fix: Codex auth reliability — persist rotated tokens, pin accounts, breaker correctness

Calmingstorm:
First of the fix series for the 2026-07-01 code review (findings confirmed by Odin's independent validation).

## Fixes

**Tier 0.2 — Codex refresh tokens destroyed on restart/reload** (the recurring "all 3 accounts die at once" outage)
- `_init_accounts` no longer unconditionally overwrites shadow files: for the same `account_id`, the newer side (by `expires_at`) wins, and a preserved shadow syncs back into the canonical entry
- `CodexAuth.on_save` hook mirrors every token rotation back to the canonical multi-account file — the persistence path that `src/setup.py` was supposed to provide but which never existed
- `account_id` identity check prevents a stale shadow from resurrecting a deleted account after `/api/codex/account/{i}` DELETE

**2.1 — Concurrent failover corrupts healthy accounts**
- New `CodexAuthPool.acquire()` → `(token, account_id, index)`; 429/401 penalize the pinned index (`mark_limited(i)` / `mark_auth_failed(i)`), not whatever account is current at lock time
- Token + account_id now an atomic pair from the same account (Tier 4 item)
- `acquire()` refreshes outside the pool lock — a slow 30s refresh no longer serializes all LLM traffic (Tier 4 item)
- All-rate-limited raises a distinct message instead of `"All 3 accounts failed: "` with an empty reason list (Tier 4 item)

**2.2 — Breaker double-counts terminal 429/5xx**
- The text and tool retry loops were duplicated copies carrying the same fall-through bug; unified into one `_send_with_retries` engine — exactly ONE `record_failure()` per failed attempt

**Reactive 401 never actually refreshes**
- `force_refresh()` exercises the refresh token immediately; `invalidate_current()` only dropped the cache and re-served the same revoked-but-unexpired bearer. Skip-if-already-rotated guard avoids concurrent double-refresh

**Stream failure/incomplete events swallowed**
- `response.failed`/`error` → `CodexStreamError` → retry/breaker path
- `response.incomplete` → partial text kept, `stop_reason="incomplete"`
- Mid-stream `asyncio.TimeoutError` (600s total) now gets breaker bookkeeping + retry parity

**Malformed tool-call args coerced to `{}`**
- Sets `ToolCall.parse_error`; both dispatchers in `client.py` return the error to the model instead of executing the tool with empty input

**Misc**
- Account-delete endpoint uses locked `reload_async()` instead of lock-ignoring sync `reload()`

## Tests
- 26 new tests (`tests/test_codex_reliability.py`): shadow preservation/write-back/identity, pinned marking, pool-lock scope during refresh, breaker single-count, timeout parity, terminal stream events, parse_error, reactive-401 e2e
- Full suite: **5757 passed, 8 skipped**

### DIFF
--- a/src/discord/client.py
+++ b/src/discord/client.py
@@ -3105,6 +3105,18 @@ class OdinBot(commands.Bot):
                 tool_name = block.name
                 tool_input = block.input
                 log.info("Tool call: %s(%s)", tool_name, tool_input)
+                # The provider could not parse the model's arguments — do NOT
+                # run the tool with a silently-empty input; bounce the error
+                # back so the model retries with valid arguments.
+                if getattr(block, "parse_error", None):
+                    log.warning("Tool call %s not executed: %s", tool_name, block.parse_error)
+                    return {
+                        "type": "tool_result", "tool_use_id": block.id,
+                        "content": (
+                            f"Error: {block.parse_error}. The tool was NOT executed — "
+                            "re-issue the call with valid JSON arguments."
+                        ),
+                    }
                 # Central RBAC gate: Discord-native tools, skills, and MCP tools are
                 # dispatched below WITHOUT going through ToolExecutor.execute() (the only
                 # place check_permission runs). permissions.filter_tools is advisory
@@ -4735,6 +4747,17 @@ class OdinBot(commands.Bot):
                 tool_name = block.name
                 tool_input = block.input
                 log.info("Loop tool call: %s(%s)", tool_name, tool_input)
+                # Provider couldn't parse the model's arguments — don't run
+                # the tool on a silently-empty input (see _run_tool).
+                if getattr(block, "parse_error", None):
+                    log.warning("Loop tool call %s not executed: %s", tool_name, block.parse_error)
+                    return {
+                        "type": "tool_result", "tool_use_id": block.id,
+                        "content": (
+                            f"Error: {block.parse_error}. The tool was NOT executed — "
+                            "re-issue the call with valid JSON arguments."
+                        ),
+                    }
 
                 t0 = time.monotonic()
                 error = None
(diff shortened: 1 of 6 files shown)


Calmingstorm:
Verdict: **REQUEST CHANGES**

The main reliability work looks sound and the targeted tests pass locally (`31 passed` across `tests/test_codex_auth_rotation.py` + `tests/test_codex_reliability.py`, `git diff --check` clean). I verified the canonical/shadow token persistence, pinned account marking, force-refresh-on-401, retry unification, stream failure handling, timeout retry path, and locked delete reload paths in the diff.

Blocking issue:

- `src/llm/openai_codex.py:724-737` still silently converts malformed tool-call arguments from the `response.completed` fallback path into `{}` with `parse_error=None`. The new parse-error handling covers `response.function_call_arguments.done` (`662-674`) and `response.output_item.done` (`685-696`), but not the completed-event fallback. That means a malformed fallback function call can still be executed with empty input, exactly the class of bug this PR says it fixed. Repro against PR head:

```python
resp = FakeResp(sse([{"type":"response.completed","response":{"output":[{"type":"function_call","call_id":"c1","name":"run_command","arguments":"{bad"}]}}]))
r = await client._read_tool_stream(resp)
assert r.tool_calls[0].input == {}
assert r.tool_calls[0].parse_error is None  # currently true, should be parse_error
```

Please apply the same `parse_error` behavior in the `response.completed` fallback and add a regression test for malformed fallback arguments.

Calmingstorm:
Re-reviewed the `45cccbe` delta. The blocker I raised is fixed.

Observed:
- `src/llm/openai_codex.py:728-739` now catches `json.JSONDecodeError` in the `response.completed` fallback path, preserves `parsed_args = {}`, and sets `parse_error = "malformed tool arguments (invalid JSON): ..."` on the `ToolCall`.
- Added focused regressions for malformed and valid completed-fallback calls in `tests/test_codex_reliability.py`.

Validation:
```text
python3 -m pytest tests/test_codex_reliability.py -q
28 passed in 0.17s
```

Verdict: approve. Ship it.

PR #124, opened 2026-07-02
fix: scheduler/agent/loop reliability — backoff, idempotent start, dedup, force-kill

Calmingstorm:
Second fix batch from the 2026-07-01 review (Odin-validated). Independent of #123 (already merged); rebased on current master.

## Fixes

**2.3 — Autonomous-loop error path tight-loops; backoff is dead code**
- Failing iterations now back off (exponential, capped) before retrying. The old `except: … continue` jumped over the wait block entirely, so a fast-failing callback hammered a down endpoint at full speed. The previously-unreachable backoff block is replaced by a shared `_interruptible_wait` used on both paths.

**2.4 — `on_ready` not idempotent → duplicate schedulers/monitors on reconnect**
- `Scheduler.start()` no-ops if a loop task is already running (refreshes callbacks only). Discord fires `on_ready` on every reconnect; this stops a second `_loop()` leaking and double-firing one-time/retry schedules.
- `InfraWatcher.start()` likewise won't spawn a duplicate set of check loops (duplicate SSH probes/alerts).

**Overlap/dedup guard**
- Per-schedule in-flight set in `_execute_and_record` drops overlapping executions (manual `run_now` during a tick, duplicate webhook deliveries).

**Shutdown abandons in-flight tasks**
- `LoopManager.shutdown()` cancels **and awaits** loop tasks; `close()` uses it instead of `stop_loop("all")` (which only set cancel events → "Task was destroyed but it is pending" + skipped trajectory save).
- `close()` awaits cancelled agent tasks so their `finally` (trajectory save) runs before exit.

**Agent kill is cooperative-only**
- `kill()`/`check_health()` now force-cancel the agent task, not just set a flag a stuck 300s tool call ignored while the log claimed "Force-killed". The run's `except CancelledError` + `finally` still finalize cleanly.

**Dead wiring**
- Scheduler `failure_callback` wired (`_on_schedule_failure`) — the consecutive-failure alert path could never fire before.
- Loop-spawned agents pass `agents.max_iterations` instead of the module default.
- `LoopManager.cleanup_finished()` is now called opportunistically and handles loops stopped before their first iteration (no `last_trigger`).

**Tier 4**
- Retry backoff off-by-one: first retry waits `base`, not `2·base`.
- `_compute_tick_delay` honors `retry_at` (retries were up to 60s late).
- Per-schedule cron timezone (IANA) via new `cron_timezone` param — plumbed through `add`/`update`, the `schedule_task`/`update_schedule` tool schemas, and `_tick` recompute. Fires on the target tz's wall clock across DST; defaults to UTC.
- `plan_id` gets a random suffix (1s-granularity collisions overwrote plans); `plans.json` written `0600` (was world-readable, embeds user requests).

## Tests
- 37 new (`tests/test_scheduler_agents_reliability.py` + fallback additions); `test_graceful_shutdown` updated for the `shutdown()` contract.
- Full suite: **5750 passed, 8 skipped**.

### DIFF
--- a/src/agents/manager.py
+++ b/src/agents/manager.py
@@ -739,7 +739,10 @@ class AgentManager:
             elapsed = now - agent.created_at
             idle = now - agent.last_activity
             if elapsed > MAX_AGENT_LIFETIME:
-                agent._cancel_event.set()
+                # Actually force-cancel the task — setting the cancel event
+                # alone left an agent stuck in a long tool call running for up
+                # to TOOL_EXEC_TIMEOUT while the log claimed "Force-killed".
+                self._force_cancel(agent)
                 killed += 1
                 log.warning(
                     "Force-killed stuck agent %s (%s): lifetime exceeded (%ds)",
--- a/src/agents/manager.py
+++ b/src/agents/manager.py
@@ -468,6 +468,22 @@ class AgentManager:
             })
         return result
 
+    @staticmethod
+    def _force_cancel(agent: AgentInfo) -> None:
+        """Signal cooperative stop AND cancel the task.
+
+        The cancel event alone is only checked between iterations, so an
+        agent stuck in a long tool call ignored a "kill" for up to
+        TOOL_EXEC_TIMEOUT (300s) — the log said "killed" while the agent ran
+        on. Cancelling the task interrupts the in-flight await; the agent
+        run's `except CancelledError` + `finally` still finalize the
+        trajectory cleanly.
+        """
+        agent._cancel_event.set()
+        task = getattr(agent, "_task", None)
+        if task is not None and not task.done():
+            task.cancel()
+
     def kill(self, agent_id: str, cascade: bool = True) -> str:
         """Cancel a running agent. If cascade=True, also kill all descendants."""
         agent = self._agents.get(agent_id)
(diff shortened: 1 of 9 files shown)


Calmingstorm:
Verdict: request changes.

Targeted validation passed:

```text
git diff --check origin/master...HEAD: clean
python3 -m pytest tests/test_scheduler_agents_reliability.py tests/test_graceful_shutdown.py -q
36 passed, 1 warning in 1.40s
```

Blocking issue: `AgentManager.check_health()` still does not force-cancel the stuck agent task.

`kill()` was correctly moved through `_force_cancel()` (`src/agents/manager.py:487-503`), which sets the cancel event and calls `task.cancel()`. But `check_health()` only sets `agent._cancel_event` when lifetime is exceeded (`src/agents/manager.py:727-754`) and never calls `_force_cancel(agent)`, so an agent stuck inside a long awaited tool call can still sit there until the tool timeout. That leaves the priority-4 “force-cancel stuck 300s tool” hole half-fixed.

Repro against PR head:

```text
result= {'killed': 1, 'stale': 0}
cancel_event= True
task_cancelled= False
task_done= False
```

Minimal fix: replace the `agent._cancel_event.set()` path in `check_health()` with `self._force_cancel(agent)` and add/adjust a regression that attaches a sleeping task and asserts it becomes cancelled/done after a loop tick.

Calmingstorm:
$Re-review of 7a30473: approved.\n\nThe blocker is fixed. `AgentManager.check_health()` now calls `_force_cancel(agent)` in the lifetime-exceeded branch, so the health-check safety net matches explicit `kill()` and actually cancels a stuck task instead of only setting the cooperative cancel event. Evidence: `src/agents/manager.py` around the health-check timeout branch now routes through `_force_cancel`.\n\nThe new regression test attaches a real sleeping task, forces the health path, and asserts the task is cancelled, covering the exact failure mode from the prior review.\n\nValidation:\n```text\npython3 -m pytest tests/test_scheduler_agents_reliability.py tests/test_graceful_shutdown.py -q\n37 passed, 1 warning in 1.20s\n```\n\nVerdict: approve.

PR #125, opened 2026-07-02
fix: session/archive persistence reliability

Calmingstorm:
Third fix batch from the 2026-07-01 review (Odin-validated). Rebased on current master (has #123, #124).

## Fixes

**2.9 — `save()` loses updates on any write failure**
- `save()` clears a channel from `_dirty` only **after** its write succeeds. The old code cleared dirty up-front, so a failed `tmp.write_text`/`replace` (disk, permission, or tmp collision between the Discord `to_thread(save)` and the web-chat save) permanently dropped that session's changes.
- `save_all()` keeps failed channels dirty for a later retry.

**2.10 — Archive write non-atomic; restore aborts on newest corrupt file**
- `_archive_session` writes tmp+rename (atomic); a crash mid-write can no longer leave a truncated newest archive.
- `_restore_from_archive` tries candidates newest-first and **falls back** past a corrupt archive instead of returning `None` on the newest (which started the channel fresh despite dozens of good older archives).

**Tier 3**
- Token estimator adds a per-message overhead (was content-only `len//4`), so the 64K send budget isn't overshot on dense code/JSON.
- Archive cap eviction protects each channel's **newest** archive — a churning channel can't evict a quiet channel's only restore point (was global oldest-by-mtime).
- Reset/purge content no longer surfaces via `search_history`: `_search_archives` and the hybrid path now filter by the channel's reset epoch, mirroring restore. (No vec-delete API needed — filter at query time.)
- Compaction feeds source messages at full length (`COMPACTION_SOURCE_MAX_CHARS = 4000`) instead of clipping to 500 before summarizing, which lost identifiers past char 500.
- Compaction/reflection `max_tokens` raised 300/500 → 1500/2000 so providers that honor the cap (Ollama/Kimi) don't truncate segments / reflection JSON to `[]`.

## Tests
- 19 new (`tests/test_session_persistence_reliability.py`); `test_token_budget` updated for the estimator overhead.
- Full suite: **5769 passed, 8 skipped**.

## Note
Excludes the RRF-merge-keys and embed-failure-backfill items (Odin marked those UNCERTAIN / they belong with the knowledge-store work in the next PR).

### DIFF
--- a/src/discord/client.py
+++ b/src/discord/client.py
@@ -674,17 +674,22 @@ class OdinBot(commands.Bot):
 
     def _wire_llm_callbacks(self) -> None:
         """Attach LLM-backed compaction and reflection callbacks using the active provider."""
+        # Compaction emits a segment of up to ~2500 chars (≈625 tokens) plus
+        # structured header lines; reflection emits multi-lesson JSON. The old
+        # 300/500 caps guaranteed mid-output truncation on providers that honor
+        # max_tokens (Ollama/Kimi) — truncated reflection JSON parsed to [] and
+        # silently dropped lessons. (Codex ignores max_tokens entirely.)
         async def _llm_compaction(messages: list[dict], system: str) -> str:
             client = self.llm_client
             if not client:
                 raise RuntimeError("No LLM provider configured")
-            return await client.chat(messages=messages, system=system, max_tokens=300)
+            return await client.chat(messages=messages, system=system, max_tokens=1500)
 
         async def _llm_reflection(messages: list[dict], system: str) -> str:
             client = self.llm_client
             if not client:
                 raise RuntimeError("No LLM provider configured")
-            return await client.chat(messages=messages, system=system, max_tokens=500)
+            return await client.chat(messages=messages, system=system, max_tokens=2000)
 
         self.sessions.set_compaction_fn(_llm_compaction)
         self.reflector.set_text_fn(_llm_reflection)
--- a/src/sessions/manager.py
+++ b/src/sessions/manager.py
@@ -61,6 +61,12 @@ TOOL_SUMMARY_THRESHOLD = 10  # summarize when this many tool calls occurred
 TOOL_SUMMARY_MAX_CHARS = 2000  # max chars for summarized tool response in history
 CHAT_RESPONSE_MAX_CHARS = 4000  # max chars for text-only (no-tool) response in history
 
+# Max chars of each source message fed into a compaction summary. Must be at
+# least CHAT_RESPONSE_MAX_CHARS or assistant responses (which persist up to
+# 4000 chars) get clipped to 500 before summarizing, permanently losing
+# identifiers past that point despite the "preserve verbatim" prompt.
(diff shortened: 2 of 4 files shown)


Calmingstorm:
Verdict: request changes.

Targeted validation passes:

```text
git diff --check origin-master...HEAD
python3 -m pytest tests/test_session_persistence_reliability.py tests/test_token_budget.py -q
51 passed in 0.44s
```

But I found two correctness holes in the reset/persistence fixes.

1. **Reset/purge content can still resurface through channel-log search.**

`SessionManager.search_history()` now filters reset epochs for archives and hybrid/vector hits, but the Step 4 channel-log fallback does not apply the channel reset epoch before appending results (`src/sessions/manager.py:1411-1437`). `_handle_purge()` calls `self.sessions.reset(channel_id)` after deleting recent Discord messages (`src/discord/client.py:3574-3580`), so content at/before that epoch is supposed to be gone from `search_history`. A channel logger hit older than `_reset_epochs[channel_id]` still comes back.

Repro against PR head:

```text
mgr._reset_epochs["c9"] = 150.0
FakeLogger.search(...) -> [{"timestamp": 100.0, "channel_id": "c9", "content": "banana old log"}]
await mgr.search_history("banana", channel_id="c9")
=> [{'type': 'user', 'content': 'banana old log', 'timestamp': 100.0, 'channel_id': 'c9', 'user_id': 'u'}]
```

2. **`save_all()` does not mark a failed clean session dirty.**

The stated behavior is “save_all() keeps failed channels dirty.” The implementation records `failed`, but then sets `_dirty = {cid for cid in self._dirty if cid in failed}` (`src/sessions/manager.py:1519-1521`). If a session was clean before shutdown/save_all and its write fails, it is not added to `_dirty`, so the failure is not retryable. This can happen after a previous successful periodic `save()` cleared dirty, then shutdown `save_all()` attempts to persist every live session and one write fails.

Repro against PR head:

```text
mgr.add_message('bad', 'user', 'x')
mgr.save()                  # dirty cleared
# make bad.tmp write fail
mgr.save_all()
=> dirty after failed save_all of clean session: set()
```

Expected: `{'bad'}`.

The fix is probably small: apply reset-epoch filtering to channel-log results in Step 4, and set `_dirty` to preserve existing dirty failures plus newly failed clean sessions, e.g. include `failed` itself rather than intersecting only with prior `_dirty`.

Calmingstorm:
Re-reviewed `831402f` delta. The two blockers from my prior review are fixed.

Evidence:
- `src/sessions/manager.py:1426-1435` now applies the reset-epoch filter to Step 4 channel-log fallback results before appending them.
- `src/sessions/manager.py:1519-1534` now assigns `self._dirty = failed`, so clean sessions that fail during shutdown save are re-marked dirty.
- Regression coverage added in `tests/test_session_persistence_reliability.py` for both paths.

Validation:
```
python3 -m pytest tests/test_session_persistence_reliability.py tests/test_token_budget.py -q
53 passed in 0.45s
```

Verdict: approve. GitHub may still reject a formal approval from this account, so posting the verdict as a comment if needed.

PR #126, opened 2026-07-02
fix: learning/knowledge reliability — memory cap, injection budget, locks

Calmingstorm:
Fourth fix batch from the 2026-07-01 review (Odin-validated). Rebased on current master (#123–125 merged).

## Fixes

**Working memory (`memory.json`) has no cap/prune**
- Per-section key cap (`MEMORY_MAX_KEYS_PER_SECTION = 200`) with LRU-by-write eviction. The full merged map is injected into every system prompt, so unbounded growth was unbounded prompt bloat.

**Learned-injection 4K budget bypassed / blown**
- The budget is now a **hard cap** on the final selection for both paths: the query-less caller (which took the include-all branch regardless of size) and the gated branch (which unioned up to 72 fixed top-K entries with no token re-check). Trim follows priority order (corrections → preferences → relevance-ranked operational → facts), always keeping at least one entry so a non-empty corpus never injects nothing.

**Consolidation fallback drops never-expire corrections/preferences**
- On consolidation-LLM failure, pin corrections and preferences instead of keeping newest-by-`updated_at` across all categories (which evicted long-standing corrections in favor of recent operational trivia).

**Reflector web `delete_entry`/`update_entry` bypass the async lock**
- New `delete_entry_async`/`update_entry_async` take `self._lock`, so a manual delete racing an in-flight reflection (lock held across a multi-second LLM await) isn't resurrected by the reflection's stale snapshot. `web/api.py` uses them.

**2.12 — Knowledge `delete_source`/`merge_sources` bypass the write lock**
- New `delete_source_async`/`merge_sources_async` take `_write_lock` and run off the event loop. `client.py` (`_handle_delete_knowledge`, now async) and `web/api.py` use them, so a delete can't interleave with an in-flight `ingest` on the shared SQLite connection (was also a sync call blocking the loop).

## Tests
- 24 new (`tests/test_learning_knowledge_reliability.py`); `test_learned_overhaul` injection tests updated for the hard cap.
- Full suite: **5792 passed, 8 skipped**.

## Note
Excludes the RRF-merge-keys and embed-failure-backfill items — Odin marked those UNCERTAIN in validation and they need a separate closer look.

### DIFF
--- a/src/discord/client.py
+++ b/src/discord/client.py
@@ -3204,7 +3204,7 @@ class OdinBot(commands.Bot):
                     elif tool_name == "list_knowledge":
                         result = self._handle_list_knowledge()
                     elif tool_name == "delete_knowledge":
-                        result = self._handle_delete_knowledge(tool_input)
+                        result = await self._handle_delete_knowledge(tool_input)
                     elif tool_name == "set_permission":
                         result = await self._handle_set_permission(
                             str(message.author.id), tool_input,
@@ -3896,7 +3896,7 @@ class OdinBot(commands.Bot):
         total = sum(s["chunks"] for s in sources)
         return f"**Knowledge base: {len(sources)} document(s), {total} total chunks**\n" + "\n".join(lines)
 
-    def _handle_delete_knowledge(self, inp: dict) -> str:
+    async def _handle_delete_knowledge(self, inp: dict) -> str:
         """Delete a document from the knowledge base."""
         if not self._knowledge_store:
             return "Knowledge base is not available."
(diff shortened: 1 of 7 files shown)


Calmingstorm:
Verdict: approve.

I reviewed PR #126 at 1345416 against master bf1c2ef.

Validation:

```text
python3 -m pytest tests/test_learned_overhaul.py tests/test_learning_knowledge_reliability.py -q
41 passed in 0.49s
```

What I checked:
- Learned prompt injection now enforces the budget on both query-less include-all and gated paths, while preserving priority order and at least one entry.
- Consolidation fallback now pins correction/preference categories instead of letting recent operational trivia evict old high-value learned entries.
- Reflector delete/update async wrappers take the same lock as reflection writes, preventing stale-snapshot resurrection.
- Knowledge delete/merge async wrappers use the store write lock and offload blocking SQLite work; Discord and Web API callers were moved to those wrappers.
- Working memory save path now caps each section at 200 keys using write-order LRU eviction.

No blocking issues found. The remaining “RRF merge” and “embed backfill” items were intentionally excluded, so I did not hold this PR on them.

PR #127, opened 2026-07-02
fix: web security + SSRF — admin gate, fail-closed auth, redaction, proxy IPs

Calmingstorm:
Fifth fix batch from the 2026-07-01 review (Odin-validated). Rebased on current master (#123–126 merged). Security-focused — worth a close read.

## Fixes

**Authorization gap on the control plane (non-admin token could self-escalate)**
- `_require_admin` guarded only ~8 of ~80 mutating routes. Rather than edit 80 routes, added a centralized `_make_admin_middleware` that enforces the **admin** tier on `ADMIN_ONLY_PREFIXES` (permissions, host-access, config, update, codex, llm, ollama, kimi, skills, mcp, tokens, personality, tool-timeouts, pools, outbound-webhooks, grafana-alerts, slack) for any method. Dev mode (no tokens configured) is unaffected, matching the existing auth middleware.
- `_require_admin` now fails **closed** on a missing identity when auth is configured (was: missing identity → allowed).

**1.1 web half — `permission_manager` attr mismatch**
- The RBAC endpoints read `getattr(bot, "permission_manager")` but the bot attribute is `permissions`, so all 4 sites always 503'd. Fixed.

**`GET /api/config` redaction leaked future secrets**
- `_redact_config` was key-name-**exact**; now substring-aware (`_is_sensitive_key`) so `hmac_key`, webhook URLs, `*_secret`, `*_password`, `private_key`, `credential` are redacted.

**2.7 SSRF**
- `http_probe`: blocks cloud-metadata endpoints (`is_metadata_url`, DNS-rebind aware) but **still allows** legitimate internal/localhost infra probing — the tool's core purpose (full private-blocking would break monitoring; the tests encode internal probing as intended). Metadata (169.254.169.254 / metadata.google.internal) is the credential-theft target that's never legitimate.
- `analyze_image` URL fetch now calls the full `is_url_blocked` (private/loopback/metadata) and disables redirect-following.

**Hardening**
- `session_timeout_minutes` defaults to **720** (was `0` = never expire → a leaked session id was valid forever).
- `web.host` bind config (was hardcoded `0.0.0.0`) + `web.trusted_proxies`.
- Rate-limit and audit use `_client_ip`: `X-Forwarded-For` honored only when the peer is a trusted proxy (else all clients behind a proxy collapse to one IP — one client could 429 everyone).
- Context loader reads with `errors="replace"` and per-file (256 KB) / total (1 MB) size caps — one undecodable byte no longer crashes boot; oversized files are skipped.

## Note on scope
`http_probe` uses **metadata-only** blocking by design (it's an infra tool). If you'd rather it block all private ranges and route internal probing through `run_command`, say so and I'll change it — but that's a behavioral change to the tool's contract, so I kept internal probing working.

## Tests
- 30 new (`tests/test_web_security_reliability.py`): admin-path matching, client-IP trust, redaction, SSRF scoping, context-loader robustness.
- Full suite: **5791 passed, 8 skipped**.

### DIFF
--- a/src/config/schema.py
+++ b/src/config/schema.py
@@ -456,8 +456,19 @@ class WebConfig(BaseModel):
     enabled: bool = True
     api_token: str = ""
     api_tokens: list[ApiTokenIdentity] = Field(default_factory=list)
-    session_timeout_minutes: int = 0
+    # Sessions expire after this many minutes of the token's lifetime. 0 meant
+    # "never expire", so a leaked WebUI session id was valid forever; default to
+    # a bounded lifetime (set to 0 explicitly to opt back into no-expiry).
+    session_timeout_minutes: int = 720  # 12 hours
     port: int = 3000
+    # Bind address. Historically hardcoded 0.0.0.0 (exposed on LAN/Tailscale);
+    # now configurable so a deployment can bind localhost and front it with a
+    # reverse proxy.
+    host: str = "0.0.0.0"
+    # Trusted reverse-proxy IPs. When the request's peer is one of these, the
+    # left-most X-Forwarded-For entry is used as the client IP for rate-limiting
+    # and audit — otherwise all clients behind the proxy collapse to one IP.
+    trusted_proxies: list[str] = Field(default_factory=list)
 
     @field_validator("port")
     @classmethod
--- a/src/context/loader.py
+++ b/src/context/loader.py
@@ -7,6 +7,12 @@ from ..odin_log import get_logger
 
 log = get_logger("context")
 
+# Per-file and total size caps. A context file is injected into EVERY LLM
+# request, so an oversized/binary file would silently bloat every prompt (or,
+# with the old unguarded read_text, crash boot on one undecodable byte).
+MAX_CONTEXT_FILE_BYTES = 256 * 1024      # 256 KB per file
+MAX_CONTEXT_TOTAL_BYTES = 1024 * 1024    # 1 MB total across all files
+
 SECRET_PATTERNS = [
     re.compile(r"(?i)(password|passwd|pwd)\s*[:=]\s*\S+"),
     re.compile(r"(?i)(api[_-]?key|apikey)\s*[:=]\s*\S+"),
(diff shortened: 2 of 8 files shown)


Calmingstorm:
Verdict: request changes.

Targeted validation passed:

```text
python3 -m pytest tests/test_web_security_reliability.py -q
30 passed in 0.43s
```

Blocker: `http_probe` still permits cloud-metadata access through redirects.

The new guard validates only the initial URL in `src/tools/http_probe_ops.py:47-67`, but `build_http_probe_command()` still enables curl redirects by default at `src/tools/http_probe_ops.py:111-115`:

```python
follow = params.get("follow_redirects", True)
if follow:
    parts.append("-L")
    parts.append("--max-redirs 10")
```

So a public URL that returns `302 Location: http://169.254.169.254/latest/meta-data/...` passes `is_metadata_url()` and then curl follows it. That bypasses the exact metadata-only SSRF boundary the PR is trying to enforce. The Norns do love leaving one door open in the wall.

Concrete proof from the current PR code:

```text
build_http_probe_command({"url": "http://example.com/redirect-to-metadata"})
=> curl ... -L --max-redirs 10 http://example.com/redirect-to-metadata
```

Expected fix: either do not follow redirects for `http_probe`, or implement redirect-chain validation so every `Location` target is checked with `is_metadata_url()` before the next hop. A regression test should cover a harmless initial URL redirecting to `169.254.169.254` and assert the probe does not reach/follow it.

I’m not blocking on the design choice to allow localhost/private probes generally. Metadata-only blocking is a reasonable operational tradeoff for this infra tool. But once metadata is declared blocked, redirect-to-metadata has to be blocked too.

Calmingstorm:
Verdict: approve.

Re-reviewed 04ab998. The redirect-to-metadata blocker is fixed.

Evidence:
- `src/tools/http_probe_ops.py` now adds `--connect-to` sinkholes for metadata endpoints whenever `follow_redirects` is enabled, covering `169.254.169.254`, `metadata.google.internal`, and `[fd00:ec2::254]` on ports 80/443 to `127.0.0.1:9`.
- The sinkhole is absent when redirects are disabled, preserving the explicit no-follow behavior.
- Regression tests cover the follow/no-follow command construction.

Validation:

```text
python3 -m pytest tests/test_http_probe_ops.py tests/test_web_security_reliability.py -q
156 passed in 0.79s
```

No blockers found in the delta.

PR #128, opened 2026-07-02
fix: tools/executor/skills reliability — RBAC wiring, skill guardrails, governor, classifier

Calmingstorm:
Sixth fix batch from the 2026-07-01 review (Odin-validated). Rebased on current master (#123–127 merged). Another security-relevant one.

## Fixes

**1.1 — RBAC gate wired into ToolExecutor**
- `PermissionManager` is now passed into `ToolExecutor`, so `check_permission()` actually enforces. It was built without it, making the "central RBAC gate" a no-op (masked live by `default_tier: admin`; a real gap for any non-admin default tier).
- `run_command` removed from `USER_TIER_TOOLS` — it's arbitrary shell, not "read-only monitoring".

**Governor/host denials reported as success**
- `_ERROR_RESULT_PREFIXES` now includes `"Blocked"` and `"Unknown or disallowed host"`, so a refused action is no longer classified `ok=True` to the LLM and audit.

**Dead code trap**
- Removed `MUTATING_TOOLS`/`READ_ONLY_TOOLS` — imported only by a schema test, consulted by no runtime authorization path. A reviewer moving a tool between them would think they'd changed authorization while nothing changed.

**`git_ops` option injection**
- `clone` inserts `--` before the URL, so `--upload-pack=...` style URLs can't be parsed as git options (shlex quoting stops shell injection, not git option injection).

**recovery / email double-send**
- `UNSAFE_TO_RETRY` += `email_send` (SMTP can fail after DATA → duplicate mail), `install_skill`, `send_to_agent`, `cancel_task`.
- Email send/search/read/list run via `asyncio.to_thread` — blocking `smtplib`/`imaplib` no longer stalls every channel.

**2.5 — Skill in-process RCE surface**
- Reject unsafe pip dependency specs (PEP 508 direct-ref / URL / VCS / path) **before** pip runs — closes the install-time `setup.py` RCE.
- `install_skill` reads the skill name **statically** (`ast.literal_eval`) instead of `exec()`ing URL code just to read a literal.
- URL AST denylist expanded (`posix`, `builtins`, `runpy`, `pdb`, `pickle`, `marshal`, …), now also catches attribute-form calls (`.system()`, `.popen()`, `__import__`) and dunder escape hatches (`__globals__`, `__subclasses__`).

**2.6 — Governor bypasses**
- `write_file` and `manage_process(write)` now go through `_govern_command` (`write_file` reached the fs via `_run_on_host`, which doesn't govern).
- Risk classifier closes the cited bypasses: `rm -rf /*` and `/`-chained, `--no-preserve-root`, `find / -delete`/`-exec rm`, `dd of=/dev/*`, `chmod 777 /` in either flag order, and `base64 --decode` / `curl|wget` piped into `(sudo) sh`. These are **escalation-only** (they make specific destructive commands stricter; scoped ops like `rm -rf /tmp/build` stay non-critical — covered by tests).

**Misc**
- `detect_mutation` covers `git_ops`, `email_send`, and `rm`/`mv`/`sed -i`/`chmod`/`chown`/`dd`/redirect commands.
- `issue_tracker` filtered from the advertised tool list when not enabled (it returned "not configured" for every call but was always offered).

## Tests
- 60 new (`tests/test_tools_executor_skills_reliability.py`). A few existing tests updated for the now-live RBAC gate (`_make_bot` uses `default_tier: admin`, mirroring live) and the tier-table change.
- Full suite: **5860 passed, 8 skipped**.

### DIFF
--- a/src/discord/client.py
+++ b/src/discord/client.py
@@ -404,12 +404,24 @@ class OdinBot(commands.Bot):
             available_hosts=list(config.tools.hosts.keys()),
         )
 
+        # Built before ToolExecutor so it can be wired in as the RBAC gate.
+        # Historically the executor was constructed without it, so
+        # check_permission() always allowed and the central RBAC gate was a
+        # no-op (masked live by default_tier: admin, but a real gap for any
+        # deployment using a non-admin default tier).
+        self.permissions = PermissionManager(
+            config_tiers=config.permissions.tiers,
+            default_tier=config.permissions.default_tier,
+            overrides_path=config.permissions.overrides_path,
+        )
+
         self.tool_executor = ToolExecutor(
             config.tools, memory_path=self._memory_path,
             browser_manager=self.browser_manager,
             output_streamer=output_streamer,
             host_access_manager=self.host_access_manager,
             email_config=getattr(config, "email", None),
+            permission_manager=self.permissions,
         )
         self.skill_manager = SkillManager(
             skills_dir="./data/skills",
@@ -489,12 +501,6 @@ class OdinBot(commands.Bot):
                 "audit.jsonl is NOT tamper-evident and verify_integrity() will fail. "
                 "Set audit.hmac_key to enable the integrity chain."
             )
-        self.permissions = PermissionManager(
-            config_tiers=config.permissions.tiers,
-            default_tier=config.permissions.default_tier,
-            overrides_path=config.permissions.overrides_path,
-        )
-
         from ..permissions.token_manager import ApiTokenManager
         self.api_token_manager = ApiTokenManager("./data/api_tokens.json")
 
@@ -1212,6 +1218,12 @@ class OdinBot(commands.Bot):
         _email_tools = {"email_send", "email_search", "email_read", "email_list_recent"}
         if not getattr(self.config, "email", None) or not self.config.email.enabled:
             builtin = [t for t in builtin if t["name"] not in _email_tools]
+        # issue_tracker returns "not configured" for every call unless enabled,
+        # yet was always advertised — so the model kept trying it. Filter it out
+        # like the other backend-gated tools.
+        issue_cfg = getattr(self.config, "issue_tracker", None)
+        if not issue_cfg or not issue_cfg.enabled:
+            builtin = [t for t in builtin if t["name"] != "issue_tracker"]
         builtin_names = {t["name"] for t in builtin}
         skill_defs = [
             t for t in self.skill_manager.get_tool_definitions()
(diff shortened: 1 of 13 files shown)


Calmingstorm:
Verdict: request changes.

Targeted validation passes:

```text
python3 -m pytest tests/test_tools_executor_skills_reliability.py tests/test_executor_integration_smoke.py tests/test_registry_schema.py tests/test_round30_reviewer.py -q
194 passed, 2 warnings in 1.14s
```

Blocker: `run_command_multi` still reports all-preflight denial cases as success.

The PR correctly classifies plain string results starting with `Blocked` / `Unknown or disallowed host` as errors, but `run_command_multi` wraps per-host denials inside markdown sections:

```text
### localhost
```
Blocked [critical]: recursive delete on root glob
```
```

That aggregate string does not start with the error prefixes, and `_handle_run_command_multi()` returns only a string, not `(output, exit_code)`, so `ToolExecutor.execute()` returns `ok=True` even when no host ran and every host was blocked/invalid.

Repro on this branch:

```python
cfg = ToolsConfig(hosts={'localhost': ToolHost(address='127.0.0.1', ssh_user='root', os='linux')})
ex = ToolExecutor(config=cfg)

await ex.execute('run_command_multi', {'hosts':['missing'], 'command':'echo hi'}, user_id='u')
# ok=True, error=None, output contains Unknown or disallowed host

await ex.execute('run_command_multi', {'hosts':['localhost'], 'command':'rm -rf /*'}, user_id='u')
# ok=True, error=None, output contains Blocked [critical]
```

This undercuts the stated fix: governor/host denials are still classified as successful for the multi-host path. Suggested fix: have `_handle_run_command_multi()` return a nonzero exit tuple when `allowed_hosts` is empty and `blocked_hosts` is non-empty, and ideally mark partial success with a warning or structured indication while keeping `ok=True` only if at least one host actually ran. Add regressions for all-blocked/all-invalid multi-host calls.

I did not find blockers in the RBAC wiring, skill URL/dependency guardrails, git clone `--`, email threading, mutation detection, or the new escalation-only classifier patterns.

Calmingstorm:
Re-reviewed ddbdedf. The run_command_multi classification blocker is fixed. The handler now returns (aggregate, exit_code), marks preflight denials, per-host errors, and no-host-ran cases as exit_code=1, and execute() now reports the previously failing missing-host and governor-blocked repros as ok=False. Targeted validation passed: python3 -m pytest tests/test_tools_executor_skills_reliability.py tests/test_executor_integration_smoke.py tests/test_registry_schema.py tests/test_round30_reviewer.py -q => 196 passed, 2 warnings. Verdict: approve.

PR #129, opened 2026-07-02
fix: discord loop / audit / config reliability (Tier 3)

Calmingstorm:
Seventh and final fix batch from the 2026-07-01 review (Odin-validated). Rebased on current master (#123–128 merged).

## Fixes

**2.11 — Audit log: no retention, uncapped input, whole-file reads**
- `tool_input` is size-capped (a full `write_file` payload was stored uncapped — 68 KB lines).
- Size-based rotation (`audit.jsonl → .1 → .2 …`, keeping `max_files`) bounds growth that had hit 48 MB / 65k lines with no retention.
- The four hot-path `search*` methods stream line-by-line into a bounded deque and read across rotated files, instead of `readlines()`-ing the whole file into RAM per query.

**2.8 — Chat tool-loop gate parity**
- Gate on `not llm_resp.tool_calls` (matching the sibling loop) instead of `not is_tool_use`, so a `stop_reason=="tool_use"` response with zero parsed calls can't skip finalization and re-loop.

**Loop cap-hit reported stale text as success**
- The loop kept `response.text` even on turns that also made tool calls, then on cap exhaustion returned that stale pre-tool text with `is_error=False` — making the cap warning unreachable. Now only a natural (tool-free) exit returns `final_text`; cap exhaustion surfaces the cap message (partial text appended).

**Concurrency / correctness**
- `_cleanup_stale_caches` no longer deletes a channel lock that is currently **held** (an in-flight request in a just-reset channel kept its lock; dropping it let the next message `setdefault` a fresh lock → two handlers in one channel).
- Orphaned user turn on cancellation: `on_message`'s `Exception` handlers miss `asyncio.CancelledError` (a `BaseException`); added a handler that cleans up synchronously and re-raises.
- Bot-message buffering uses `channel_config.should_respond_to_bots` (per-channel override), agreeing with the mention gate — a channel that opts out no longer gets its bot messages buffered and answered.
- Secret-scrub message-delete handles `NotFound`/`HTTPException` (not just `Forbidden`), so a rate-limited/already-deleted message no longer propagates out of `on_message` and drops the scrub notice.

**Config / wiring**
- `load_config` warns on unknown top-level keys (typos were silently dropped) — a warning, not `extra="forbid"`, so a slightly-ahead config can't brick boot.
- `graceful_degradation` thresholds are passed into `SubsystemGuard` (built with no args → config ignored).

**Docs**
- README tool counts corrected to 74 (were 70/72) + email mention.

## Tests
- 10 new (`tests/test_audit_config_reliability.py`): rotation, input cap, streaming across rotation, config warning, guard thresholds.
- Full suite: **5904 passed, 8 skipped**.

## Deferred (flagged, not silently dropped)
Lower-severity display/delivery polish I did **not** include, to keep monolith-edit risk down for a final PR — happy to do these as a follow-up if you want them now:
- Code-fence chunking boundary corruption
- `_send_chunked` middle-chunk-drop handling
- `_pending_files` delivery in loop/agent/scheduled contexts
- Trigger-cog `set_config`/`set_scheduler` wiring (was MOSTLY CONFIRMED — needs a verify pass)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 An autonomous execution agent on Discord. Norse god of wisdom and war, stuck managing mortal infrastructure for eternity.
 
-Odin executes real work from Discord: incident response, deploys, investigations, code review, automation, and scheduled operations across 70 tools (25 core, 47 skill). It runs shell commands on managed hosts, uses browser automation, orchestrates agents and workflows, and verifies results automatically after service changes.
+Odin executes real work from Discord: incident response, deploys, investigations, code review, automation, and scheduled operations across 74 built-in tools plus user-created skills. It runs shell commands on managed hosts, uses browser automation, orchestrates agents and workflows, and verifies results automatically after service changes.
 
 ## Why operators pick Odin
 
@@ -70,7 +70,7 @@ Tell Odin things like:
 ```
 Discord ──> OdinBot (client.py)
                │
-               ├── Tool Executor ──> 70 tools (shell, browser, git, docker, etc.)
+               ├── Tool Executor ──> 74 tools (shell, browser, git, docker, email, etc.)
                │       │
                │       ├── CommandGovernor (blocks dangerous commands)
                │       ├── Risk Classifier (observability tags)
@@ -212,7 +212,7 @@ src/
     cogs/                  9 moderation/utility cog extensions
   tools/
     executor.py            Tool dispatch + recovery
-    registry.py            72 tool definitions (25 core, 47 skill)
+    registry.py            74 built-in tool definitions
     skill_manager.py       Skill CRUD + AST validation
     risk_classifier.py    CommandGovernor + risk classification
     post_validation.py     validate_action implementation
--- a/src/audit/logger.py
+++ b/src/audit/logger.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import json
+import os
+from collections import deque
 from collections.abc import Callable
 from datetime import datetime, timezone
 from pathlib import Path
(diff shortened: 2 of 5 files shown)


Calmingstorm:
Verdict: request changes.

Targeted validation passes:

```text
python3 -m pytest tests/test_audit_config_reliability.py tests/test_client.py tests/test_stuck_loop.py tests/test_tool_loop_helpers.py tests/test_config.py -q
141 passed, 1 warning in 0.80s
```

Blocker: audit rotation breaks HMAC-signed audit logging.

`AuditLogger._maybe_rotate()` rotates the current file but the signer state is not reset and `verify_integrity()` still verifies only the new current `audit.jsonl` from genesis. After rotation, the first entry in the new current file has `_prev_hmac` pointing at the last entry in `audit.jsonl.1`, so integrity verification fails immediately.

Minimal repro on this PR:

```text
logger = AuditLogger(path, hmac_key='k', max_bytes=500, max_files=2)
# write enough entries to rotate
await logger.verify_integrity()
=> {'valid': False, 'total': 1, 'verified': 0, 'first_bad': 1,
    'error': 'Line 1: HMAC verification failed (tampered or reordered)'}
```

The code comment says HMAC resets at each rotation boundary, but the implementation does not reset `self._signer.prev_hmac` after rotation. Either reset the signer to genesis after a successful rotate, or teach verification/initialization to verify/read the chain across rotated files in oldest-to-newest order. The former matches the comment; the latter preserves a continuous chain. As written, enabling `audit.hmac_key` plus rotation makes `/api/audit/verify` report a broken audit log despite no tampering.

I did not find blockers in the Discord loop finalization/cap-hit changes, HELD lock cleanup, CancelledError orphan cleanup, per-channel bot-buffer gate, secret scrub exception handling, unknown top-level config warning, or graceful_degradation threshold wiring.

Calmingstorm:
Re-reviewed 0594f78. Verdict: approve.\n\nThe HMAC rotation blocker is fixed: AuditLogger now routes log_execution/log_event/log_web_action through _persist(), and _persist() rotates before signing/appending. _maybe_rotate() resets the signer prev hash to GENESIS, so the first entry in a fresh current audit.jsonl verifies from genesis while rotated files keep their own self-consistent chains.\n\nValidation run:\n\n```\npython3 -m pytest tests/test_audit_config_reliability.py tests/test_client.py tests/test_stuck_loop.py tests/test_tool_loop_helpers.py tests/test_config.py -q\n142 passed, 1 warning in 0.61s\n```\n\nI also manually reproduced rotation with HMAC enabled, verified current and rotated files individually, then simulated restart + append via initialize_chain(); verify_integrity() stayed valid. No remaining blockers found.

PR #130, opened 2026-07-02
fix: make ok=False tool results visible to the model

Calmingstorm:
Follow-up from Odin's post-deploy v3.44.0 smoke test (test 2 FAIL).

**Gap:** `run_command_multi`'s all-denied aggregate carries `ok=False` after #128 — audit and the scheduler path classify it correctly — but the model only ever sees `str(ToolResult)`, which is the bare output. The markdown-wrapped denial has no error prefix, so the *model* still read a refused action as success.

**Fix:** `OdinBot._ensure_failure_visible(text, ok)` — when `ok=False` and the text doesn't already start with an `_ERROR_RESULT_PREFIXES` marker, prefix it with `Error (tool reported failure):`. Wired into both dispatchers (`_run_tool`, `_run_loop_tool`); both also propagate the failure into the audit `error` field when it would otherwise be empty. Results already carrying an error marker are not double-prefixed.

**Tests:** 4 new (`tests/test_tool_failure_visibility.py`), including an end-to-end `run_command_multi` unknown-host denial through the real executor. Full suite: **5909 passed**.

### DIFF
--- a/src/discord/client.py
+++ b/src/discord/client.py
@@ -36,6 +36,7 @@ from ..scheduler import Scheduler
 from ..sessions import SessionManager
 from ..sessions.manager import CHAT_RESPONSE_MAX_CHARS, summarize_tool_response
 from ..tools import ToolExecutor, SkillManager, ToolResult, get_tool_definitions
+from ..tools.executor import _ERROR_RESULT_PREFIXES
 from ..search import LocalEmbedder, SessionVectorStore
 from ..permissions import PermissionManager
 from ..permissions.host_access import HostAccessManager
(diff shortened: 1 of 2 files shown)


Calmingstorm:
Reviewed PR #130 at 0d640db. Verdict: LGTM. The fix addresses the actual model-facing failure: ToolResult.ok=False was structurally visible to audit/scheduler but not necessarily to the model because str(ToolResult) can be clean markdown. _ensure_failure_visible() is narrowly scoped, preserves already-marked failures, and is wired through both normal tool execution and loop tool execution. Audit now also gets an error field for ok=False/no-error cases. Validation run: python3 -m pytest tests/test_tool_failure_visibility.py -q => 4 passed; python3 -m pytest tests/test_tools_executor_skills_reliability.py tests/test_executor_integration_smoke.py tests/test_registry_schema.py tests/test_round30_reviewer.py -q => 196 passed. No blockers found.

PR #132, opened 2026-07-05
refactor(client): P1 — composition root (wiring.py)

Calmingstorm:
## RFC-001 Phase 1 — Composition root

Extracts subsystem construction and teardown from `OdinBot` into `src/discord/wiring.py`. Pure move phase: **zero behavior change intended**, proven by the P0 characterization suite passing unmodified.

### What moved
- `build_services(config) → BotServices` — all ~35 bot-independent subsystems, constructed in the exact historical order (trajectory savers → loop bridge, permissions/host-access → executor, search stores → sessions, scheduler → `skill_manager.set_services`). `OdinBot.__init__` drops from 455 to ~150 lines: state fields + `build_services` + flat attribute attachment + bot-coupled construction.
- `shutdown_services(bot)` — the `close()` teardown sequence, verbatim: same order, same getattr guards (including the late-bound `health_server`/`process_registry` and the `knowledge` property alias). `OdinBot.close()` is now 4 lines.

### What deliberately did NOT move
- The ~25 per-channel mutable state fields (**P2**: ChannelStateRegistry)
- `_llm_provider_lock`/`_llm_active_requests`/`_llm_switching` (**P4**: LLMGateway)
- `VoiceManager` (constructor takes the bot), `InfraWatcher` (alert callback is a bot method), `_wire_llm_callbacks`, prompt build, command registration — all still in `__init__`, marked with their destination phases.

### Ordering deltas (all inert — nothing starts running during construction)
1. `_wire_llm_callbacks` now runs after scheduler/audit construction (it only touches `sessions`/`reflector`, both already built).
2. `voice_manager`/`infra_watcher` construct after the build-loop modules instead of before.
3. Personality preset registration moved to the end of `build_services` (still before the first `_build_system_prompt`).

One token normalized: `asyncio.TimeoutError` → `TimeoutError` in the moved teardown (builtin alias since 3.11) so the new file is ruff-clean.

### Verification
- Characterization suite: **127 passed, unmodified** — including the facade contract (all Appendix B attributes still present, late-bound names still absent) and graceful-shutdown behavior.
- Full suite: **6040 passed / 4 skipped**, runtime unchanged.
- Ruff: `wiring.py` fully clean; `client.py` has strictly FEWER findings than baseline (pre-existing E501/UP041/I001 debt untouched — no drive-by cleanups), zero new findings.
- `client.py`: 5,790 → 5,372 lines.

### Revert
Single revert of this merge restores the P0 state exactly.

### DIFF
--- a/src/discord/client.py
+++ b/src/discord/client.py
@@ -19,9 +19,7 @@ from ..monitoring import InfraWatcher
 from .background_task import (
     BackgroundTask, run_background_task, create_task_id, MAX_STEPS,
 )
-from ..agents import AgentManager, LoopAgentBridge
 from ..agents.manager import AGENT_BLOCKED_TOOLS, filter_agent_tools
-from ..tools.autonomous_loop import LoopManager
 from ..llm import CircuitOpenError, CodexChatClient, KimiClient, OllamaClient
 from ..llm.codex_auth import CodexAuthPool
 from ..llm.secret_scrubber import scrub_output_secrets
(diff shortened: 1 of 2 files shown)


Calmingstorm:
Verdict: approve with one small cleanup before merge.

I checked PR #132 against `refactor/client-decomposition` and focused on the P1 contract: construction order, teardown equivalence, intentional leftovers in `__init__`, and the P0 facade/characterization guarantees.

Validation run locally:

```text
git diff --check refactor/client-decomposition...HEAD
python3 -m pytest tests/characterization tests/test_client.py tests/test_graceful_shutdown.py -q
# 148 passed, 1 warning
ruff check src/discord/wiring.py tests/characterization tests/fakes -q
# clean
```

What looks good:
- `build_services(config)` preserves the real service construction order for the moved subsystem block, including scheduler before `skill_manager.set_services(...)` and the search/session/channel wiring dependencies.
- The three PR-body ordering deltas look inert on current code: `_wire_llm_callbacks()` still only depends on already-built session/reflector state; voice/infra remain bot-coupled; user preset registration still precedes first prompt build.
- `shutdown_services(bot)` is equivalent to old `close()` teardown after accounting for `self` → `bot`, the docstring/log/super-close wrapper split, and `asyncio.TimeoutError` → `TimeoutError`. The getattr guards and teardown sequence are preserved.
- The things that should remain bot-owned for later phases mostly do: channel state/caches, LLM lock trio, voice manager, infra watcher, prompt/command registration.
- Facade attachment is flat, and the P0 contract/characterization suite passes unmodified in the focused run.

Finding to fix before merge:

1. `OdinBot.__init__` still constructs the agent/loop/trajectory services before calling `build_services()`, then immediately overwrites them with the services returned from wiring.

   Current lines:
   - `src/discord/client.py:261-281` constructs `AgentManager`, `LoopManager`, `TrajectorySaver`, `AgentTrajectorySaver`, `LoopAgentBridge`, and `LoopReflectionGate`.
   - `src/discord/client.py:301-328` then assigns the `build_services()` instances over those same attributes.

   This is harmless-looking today because those constructors mostly allocate dictionaries / mkdir existing trajectory dirs, but it is not a pure move anymore: those subsystems are initialized twice, with extra filesystem side effects from the trajectory savers and dead transient managers. It also contradicts the PR body’s construction-order story because the first throwaway construction still happens in the old `__init__` location.

   Requested change: delete the leftover block at `client.py:261-281` and the now-unused imports (`AgentManager`, `LoopAgentBridge`, `LoopManager`) from `client.py` unless they are needed elsewhere after that removal.

After that cleanup, I’m comfortable calling P1 good. No other blockers found.

PR #159, opened 2026-07-06
ci: PR test gate — pytest + no-new-lint-findings diff (RFC-003 P0)

Calmingstorm:
## RFC-003 P0 — the safety net becomes law

Two jobs on every PR (+ report-style on master pushes):

- **tests** — `pip install -e ".[dev]"` → `pytest -q`, Python 3.12, pip cached. The 4 standing skips are version-conditional and identical on a bare runner (verified).
- **lint-no-new** — `scripts/ci/lint_gate.py` (committed, stdlib-only): same-binary ruff over merge-base and HEAD via `git worktree`, findings normalized to (file, code, source-line-text) so pure line shifts can't false-positive, **fail only on NEW findings** (resolved ones get thanked). `fetch-depth: 0` per R1 so merge-base has history. Self-consistent across ruff releases: both sides of the diff always run the same binary.

### Local validation (real history, not synthetic only)
- base = v3.45.0 tag vs HEAD → `new=0, resolved=11` — reproduces the RFC-002 lint ledger exactly.
- synthetic `import os, sys` appended to a src file → exit 1 with all 5 findings printed.
- auto-merge-base path (no args) resolves against `origin/master` → clean.

### The proof is this PR itself
Both jobs run on this very PR (pull_request workflows execute from the PR merge ref) — check the checks tab: green here means the gate gates.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,44 @@
+name: Tests
+
+on:
+  pull_request:
+  push:
+    branches: [master]
+
+concurrency:
+  group: tests-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+          cache: pip
+          cache-dependency-path: pyproject.toml
+      - name: Install
+        run: pip install -e ".[dev]"
+      - name: Run the suite
+        run: pytest -q
+
+  lint-no-new:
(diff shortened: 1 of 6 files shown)


Calmingstorm:
LGTM. Reviewed RFC-003 P0 CI gate at ec5e9cd against base refactor/api-decomposition. Verified test.yml has PR/push pytest gate and fetch-depth: 0 lint job; lint_gate.py compares same-ruff finding multisets via worktree/merge-base with normalized file/code/source-line findings; fresh-boot monotonic sentinel fixes are correct and regression-covered. Targeted regression tests passed, lint gate clean vs PR base, GitHub checks are green. No findings. GitHub would not let this account submit a formal approval because it sees it as the PR author, because of course the forge has opinions.

PR #161, opened 2026-07-06
refactor(web): P2 — api package swap + carve wave 1 (llm_admin, security, observability)

Calmingstorm:
## RFC-003 P2 — the swap and the first 25 sections

**Swap**: `src/web/api.py` → `src/web/api/__init__.py` verbatim, relative imports depth-bumped (module-level AND the indented late imports inside handlers — the second category initially escaped and the web tests caught it). The P1 import-surface pin guarded the whole move.

**Carve**: 25 sections → three domain modules as **per-section registrars called at each section's original position** — registration order preserved exactly, proven by the parity contract's order check. `__init__` is down to 2,406 lines; wave-1 modules: `llm_admin.py` 875 / `security.py` ~500 / `observability.py` ~400.

**Severed closures, rehomed** (declared deviations from verbatim):
- `admin_gate(bot)` factory in `api_common` — one canonical `_require_admin`; composition root + tokens registrar both build from it.
- `_codex_creds_lock` → module-level in `api_common` (one per process ≡ old one-per-table).
- The provider-config cluster hoisted to `llm_admin` module scope; `_persist_config` pair takes `bot` explicitly (6 call sites).

**The incident your review should savor**: carve pass 2 swallowed the already-inserted `register_aggregates` call into the moved `usage_cost` block. The aggregates routes **still registered** — from inside the wrong registrar — so the parity count stayed 183 and every test stayed green. An import-vs-call audit caught it; the call is back at its original position. Lesson encoded: green parity ≠ clean composition; the wave-2+ carve script will assert calls == defs.

**Lint**: the rename makes every pre-existing api.py finding 'new' to the gate → P6's burn-down arrived early for these files, all fixed: **gate new=0, resolved=58**. Suite **6,079 passed**. CI green on this PR.

### DIFF
--- a/src/web/api/__init__.py
+++ b/src/web/api/__init__.py
@@ -15,21 +15,20 @@ from typing import TYPE_CHECKING
 from aiohttp import web
 from croniter import croniter
 
-from ..config.schema import Config
-from ..odin_log import get_logger
-from ..setup_wizard import (
+from ...config.schema import Config
+from ...odin_log import get_logger
+from ...setup_wizard import (
     build_config,
     build_env,
     is_setup_needed,
     validate_token_format,
 )
-from ..tools.registry import get_tool_definitions
-from ..version import get_version
+from ...version import get_version
 
 # Shared helpers live in api_common (RFC-003 P1) — re-imported so existing
 # spellings (`from src.web.api import _redact_config`, internal uses) and
 # patch targets keep working through the carve.
-from .api_common import (  # noqa: F401 — re-exports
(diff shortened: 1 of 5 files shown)


Calmingstorm:
LGTM. Reviewed P2 file→package swap and wave-1 carve. Verified route parity/order remains pinned at 183 routes, R1 import surface survives the package swap, admin_gate semantics match the old _require_admin closure, _codex_creds_lock identity is shared via api_common, provider config persistence hoist passes focused tests, register_aggregates is called at the original observability position after usage-cost, and CI is green. No findings.

PR #165, opened 2026-07-06
docs: RFC-003 results + R2; README gains the web package layout

Calmingstorm:
## RFC-003 P7 (docs half)

- Results section in the plan: 4,155 → 217-line composition root, 13 domain modules + api_common (none >600), 183 routes exact-order through every wave, 48 registrars each called once, suite 6,079 throughout.
- R2 records the declared deltas: **P6 absorbed per-wave** (package rename surfaced all old api.py debt to the gate in wave 1; package finishes at zero ruff findings — no separate burn-down, no baseline re-anchor), the two size-gate splits, and the two domain-correct appends.
- README project-structure block gains the `src/web/` layout.

The soak half follows: deploy to /opt/odin + the web battery.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01ABCWoxMmuTNWWWnjJ9wrg3

PR #166, opened 2026-07-06
RFC-003: CI gate + api.py decomposition — 4,155-line monolith to composition root + 13 domain modules

Calmingstorm:
## The campaign (7 reviewed PRs, #159–#165)

Two deliverables, planned as RFC-003 (`docs/plans/api-decomposition-plan.md`, R0–R2 with results):

1. **First-ever PR CI gate** — `.github/workflows/test.yml` runs the full pytest suite plus `scripts/ci/lint_gate.py`, a no-new-findings comparator that diffs ruff results against the merge-base (normalized to file/code/stripped-line, so pre-existing debt doesn't block but new debt does). The gate's very first run caught **7 real production bugs** — the `time.monotonic()` zero-sentinel class, where boot-relative "never" reads as "recently" on a young process (grafana-alerts cooldown, autonomous-loop cleanup) — all fixed with fresh-boot regression tests.
2. **`api.py` retired** — the 4,155-line monolith is now a **217-line pure composition root** (`src/web/api/__init__.py`) plus **13 domain modules** and `api_common`, none over 600 lines. A characterization contract pins all **183 routes in exact registration order** (order is aiohttp path-precedence behavior) and held through every wave; 48 registrars are each called exactly once from the root, enforced by executable assertions; the `process_web_chat` patch seam resolves at call time so existing test patches keep governing the real handlers. The package carries **zero ruff findings**.

Every PR was bridge-reviewed by Odin before merging.

## Soak

Deployed to the live install @ `3baf754` — two soak batteries executed by Odin:

- **Round 1**: REST GET sweep, live chat through the decomposed package, one-shot schedule fire + auto-remove, skill lifecycle, memory round-trip. All green; the single flagged 405 was a battery-spec error (endpoint never existed in any version — verified against the parity table, the v3.46.0 baseline, and live probes of both).
- **Round 2**: auth-gate negative paths (401s), config redaction verified field-by-field, full REST mutation cycles (schedules, knowledge, memory) with clean teardown, 21-endpoint status sweep across every domain module, 404 semantics, session persistence. Zero regressions; the flagged items were all confirmed honest subsystem state or pre-existing conventions.

Zero restarts, zero campaign errors, RSS flat vs baseline across the whole soak.

## Gates

6,079 passed · route parity 183 exact-order · lint zero-new · `ui/` untouched

Merging with a merge commit to preserve per-PR history.


Calmingstorm:
LGTM. Reviewed the merged RFC-003 integration whole against master: CI gate, route/import parity, registrar composition, package split, and targeted validation all check out. No findings.

PR #169, opened 2026-07-06
RFC-004 P2: explicit late-bound dispatch table over unmoved handlers

Calmingstorm:
## RFC-004 P2 — late-bound dispatch table (behavior-identical wave)

`EXECUTOR_HANDLERS: dict[str, tuple[str, str]]` — 28 explicit `name → (owner_key, attr)` entries, all pointing at the `"core"` owner until the P4–P6 waves rebind them to domain owners. Resolution is `getattr(owner, attr)` at **call time** in `_resolve_handler`, never pre-bound — the shape your plan-review blocker #1 demanded, mirroring the native dispatcher.

- Legacy f-string getattr retained as a logged fallback (P7 retires it)
- Resolution tolerates `__init__`-bypassing fixtures (`ToolExecutor.__new__` in test_http_probe_ops) by falling through to the legacy path — identical behavior while every entry is core-bound
- **+4 contract tests**: table keys == executor-routed set · every owner_key bound · per-entry table-vs-legacy resolution identity (the migration bridge assertion) · patch-seam THROUGH the table

Full suite: **6,123 passed** · lint gate vs campaign base: **new=0** (the 50 pre-existing executor.py findings are untouched baseline, absorbed when the core slims in P4–P6)

### DIFF
--- a/src/tools/executor.py
+++ b/src/tools/executor.py
@@ -106,6 +106,44 @@ def _build_bulkhead_registry(config: ToolsConfig) -> BulkheadRegistry:
     return registry
 
 
+# RFC-004 P2: explicit late-bound dispatch table — tool name -> (owner_key, attr).
+# Handlers are resolved via getattr(owner, attr) at CALL time, never pre-bound,
+# so instance-attribute patches (``executor._handle_x = fake``) keep governing
+# execution (the patch-seam contract in test_executor_dispatch_parity). All
+# entries point at the "core" owner until the P4–P6 waves rebind them to domain
+# owners. The characterization contract pins table keys == executor-routed set.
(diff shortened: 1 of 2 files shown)


Calmingstorm:
LGTM. Reviewed RFC-004 P2 at b512ff6 against refactor/tools-decomposition. Verified the 28-entry EXECUTOR_HANDLERS table is core-bound, late-bound via getattr at call time, matches the executor-routed set, and preserves the table-vs-legacy identity bridge. The patch seam holds through execute() with the table path, context/middleware pins remain green, __init__-bypass tolerance falls back to legacy resolution, lint gate reports new=0, and GitHub checks are green. No findings.

PR #170, opened 2026-07-06
RFC-004 P3: skill dispatch extracted to native_tools/skills_tools.py

Calmingstorm:
## RFC-004 P3 — skill dispatch extracted from the native table

The ~120 lines of inline skill `if/elif` (CRUD, meta, invoke_skill, dynamic user skills, and the message/file callback factories) move **verbatim** from `NativeToolDispatcher.dispatch()` to a `SkillTools` domain owner — the hidden 6th domain RFC-002 left embedded in the dispatch-table file.

- `SkillTools` is built by the dispatcher on the **same objects** (shared references) — monkeypatching `dispatcher.skill_manager.has_skill` reaches both, so the P0 truth-table pins hold unmodified
- `effects` is constructed by the registry and passed INTO `skills.dispatch(...)` — skills_tools never imports registry, no cycle
- `dispatcher.handles()` = table check + `skills.handles()` — identical truth table
- `SKILL_CRUD_TOOLS` re-exported from registry for existing importers
- The two file-callback pins re-pointed at `d.skills._skill_file_cb` (the per-wave re-pointing rule from the plan); every other pin passes untouched
- native `registry.py` 324 → 193 lines · both files at **zero** ruff findings

Full suite: **6,123 passed** · characterization: 188 passed

### DIFF
--- a/src/discord/native_tools/registry.py
+++ b/src/discord/native_tools/registry.py
@@ -32,14 +32,14 @@ Behavioral contract (pinned by the P0 characterization suite):
 from __future__ import annotations
 
 import asyncio
-import io
 from dataclasses import dataclass
 from typing import Any, Literal
 
-import discord
-
 from ...odin_log import get_logger
-from ..response_guards import scrub_response_secrets
+from .skills_tools import (  # noqa: F401 — re-export (P0 pins import SKILL_CRUD_TOOLS from here)
+    SKILL_CRUD_TOOLS,
+    SkillTools,
+)
 
 log = get_logger("discord")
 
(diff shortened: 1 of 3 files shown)


Calmingstorm:
LGTM. Reviewed RFC-004 P3 skill extraction at 0e23d95 against refactor/tools-decomposition. Verified P0 skill/native dispatch pins, effects-passed-in seam/no registry import cycle, file-callback pins re-pointed to d.skills, and AST/statement identity for the moved skill CRUD/meta/invoke/dynamic dispatch bodies. Local contracts: 44 passed; ruff touched files clean; lint gate new=0. CI checks are green.

PR #177, opened 2026-07-06
Hardening: nonzero exit on fatal startup + audit HMAC unsigned-prefix tolerance

Calmingstorm:
Two small operational-hardening changes, one commit each. Both designs were discussed with Odin over the bridge before implementation; all his advisories are incorporated.

## 1. `main: exit nonzero on fatal startup`

Fatal startup errors (bad token, boot-time DNS failure) were logged and swallowed — `main()` returned normally, exiting **0**. Failures outside `run()`'s guard (e.g. health-server port bind) tracebacked out with no clean shutdown. Supervisors could never distinguish a crash from a clean stop; the packaged unit only recovers because it uses `Restart=always`.

- `run()`'s fatal path now sets exit code 1; `main()` ends with `sys.exit(code)` only when nonzero (no unconditional `sys.exit(0)` churn).
- Failures escaping `run()` get logged, a guarded cleanup (cleanup failures are logged separately and never mask the fatal code), and exit 1.
- Intentional `SystemExit(N)` keeps its original code — never normalized to 1.
- Clean stops stay exit 0: SIGTERM, Ctrl-C, top-level `CancelledError` (BaseException, so it can't be caught by the fatal `except Exception` — pinned by test), close-triggered return.
- New `tests/test_main_exit_codes.py` (8 tests) drives the real `main()` control flow with faked `OdinBot`/`HealthServer`/`load_config`, asserting both exit codes and that cleanup ran.

## 2. `audit: tolerate unsigned prefix when HMAC is enabled`

`verify_log()` failed the **entire file** at the first entry missing `_hmac`, so enabling `audit.hmac_key` on an install with existing unsigned history left `GET /api/audit/verify` permanently 409. Enabling tamper-evidence should not require discarding forensic history.

Invariant: `unsigned* signed+` is valid; `signed+ unsigned` is not.

- Entries predating enablement are tolerated as an *unsigned prefix*, counted in a new `unsigned_prefix` response field (present in every verify response, including errors and not-enabled, for a stable shape).
- The chain is verified strictly from the first signed entry, which chains from `GENESIS` — exactly what `initialize_chain()` produces over an all-unsigned file. Strict tamper semantics after that point are unchanged (missing/bad `_hmac`, bad prev, reorder all still fail).
- `verified` now unambiguously counts signed entries only. Verification stays per-file; rotated files keep their own self-consistent chains.
- Rejected alternative: auto-rotate-on-enable (pushes pre-HMAC history into the `.1`–`.5` rotation cycle where it eventually gets deleted, and hides the provenance boundary instead of representing it).
- Docs: config template explains key generation and the transition semantics; the startup warning mentions enabling-later is safe.
- Tests: the strict pin is deliberately updated with a migration-semantics comment; new `TestUnsignedPrefix` covers all-unsigned, prefix+valid-chain, unsigned-after-signed (tamper), bad-HMAC-after-prefix, zero-prefix shape; a new endpoint test exercises the full enablement flow (existing unsigned file → key on → `initialize_chain()` → new signed entry → 200 with counts).

## Verification

- Full suite: **6,144 passed, 4 skipped** (baseline 6,133 + 15 new)
- Lint gate vs merge-base: **new=0**

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -144,6 +144,10 @@ attachments:
   retention_hours: 24
 
 # Audit log integrity — HMAC signing for audit.jsonl entries
+# Generate a key with: openssl rand -hex 32
+# Enabling on an existing install is safe: history written before enablement
+# is not retroactively signed — /api/audit/verify reports it as
+# unsigned_prefix and verifies the chain from the first signed entry on.
 audit:
   hmac_key: ''  # Set a secret to enable HMAC-SHA256 signing; empty = disabled
 
--- a/src/__main__.py
+++ b/src/__main__.py
@@ -130,7 +130,14 @@ def main() -> None:
     loop = asyncio.new_event_loop()
     asyncio.set_event_loop(loop)
 
+    # Fatal paths set this nonzero so supervisors (systemd Restart=on-failure,
+    # monitoring) can distinguish a crash from a clean stop. Historically every
+    # fatal startup error (bad token, boot-time DNS failure) exited 0 and only
+    # Restart=always kept the service recovering.
+    exit_code = 0
+
     async def run() -> None:
+        nonlocal exit_code
         await health.start()
 
         async def _webhook_send(channel_id: str, text: str) -> None:
@@ -158,6 +165,7 @@ def main() -> None:
             log.info("Connecting to Discord…")
             await bot.start(config.discord.token)
         except Exception as exc:
+            exit_code = 1
             log.error("Fatal error: %s", exc, exc_info=True)
             await shutdown()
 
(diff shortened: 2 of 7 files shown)


Calmingstorm:
LGTM. I reviewed PR #177 at `e8b4def` against `master` (`13e6e4b2608f`).

Observed:
- Startup fatal paths now set a nonzero `exit_code` and exit via `sys.exit(exit_code)` only when needed.
- `bot.start()` `Exception` paths log, clean up, and exit 1.
- outer `run_until_complete(run())` guard handles pre-inner-try failures such as `health.start()`/bind errors, logs cleanup separately, and exits 1.
- `KeyboardInterrupt` and top-level `asyncio.CancelledError` remain clean stops.
- intentional `SystemExit` code is preserved.
- audit verification now accepts only an unsigned prefix before the first signed entry, reports `unsigned_prefix`, and remains strict once the signed chain begins.
- all-unsigned logs with HMAC enabled verify as valid with `verified=0` and `unsigned_prefix=N`, matching the enablement migration semantics.
- docs/warnings explain that enabling HMAC does not retroactively sign old history.

Validation I ran:
```text
OK

python3 -m pytest tests/test_main_exit_codes.py tests/test_audit_signing.py -q
101 passed, 1 warning in 0.83s

GitHub checks:
tests        COMPLETED SUCCESS
lint-no-new  COMPLETED SUCCESS
```

No blockers found. The two commits are cleanly scoped and incorporate the prior advisories.

PR #178, opened 2026-07-06
learning: repair damaged entries during consolidation

Calmingstorm:
Closes the learned-memory dead end: entries clipped over the 1,000-char cap were flagged `damaged: true` "until resummarized," but no resummarization existed — they stayed quarantined forever, injecting truncation-marker noise and permanently shrinking the consolidation target (`target = consolidation_target − len(damaged)`).

Design was discussed with Odin over the bridge before implementation; all advisories incorporated.

## What changes

`_consolidate` now runs a **repair pass** on damaged entries after the damaged/candidates partition and before the merge:

- Per-entry `_text_fn` resummarize, budget **5 per cycle** (historical damaged counts: 0–4), with a strict preserve-meaning prompt ("rewrite only the given lesson; do not infer missing details; no examples/causes/context not present").
- The truncation marker is stripped from the LLM input; output must survive a **validation gauntlet**: non-empty · not JSON-shaped (`]`/`}` count as sentence terminators, so serialized output would otherwise pass the chopped heuristic) · no truncation marker (rejects marker echoes) · ≤ hard cap · not `_looks_chopped` · clean through `_clip_content`.
- **Success**: content replaced, flag cleared, entry rejoins the candidate pool — target math uses the *remaining* damaged count, so each repair reclaims its consolidation slot immediately.
- **Any failure** (no backend, LLM error, rejected output, budget spent): the original entry object passes through untouched, in order — byte-for-byte the previous passthrough behavior.
- Only `content` and `damaged` are ever modified. Repair is maintenance, not reuse: timestamps and expiry semantics unchanged. No schema/config/API changes.

## Tests (+8, one deliberate pin update)

New `TestDamagedRepair`: success clears flag+marker · unrelated fields (incl. all timestamps) unmutated · marker stripped from repair input · per-entry exception isolation (entry 2 failing doesn't stop 3–5) · budget 7→5 attempted, 2 untouched in order · rejection matrix (empty / >1000 / chopped / marker-echo prompt-injection / JSON-shaped) leaves same objects untouched · no-text_fn passthrough · target-math regression (3 damaged, 2 repaired → merge target `T−1`, not `T−3`, asserted from the captured merge prompt, which also proves clean lessons never reach repair and unrepaired entries never reach the merge).

The pin `test_damaged_entries_bypass_the_llm` is deliberately updated (with comment): damaged entries now reach the LLM **for repair only**, and still never enter the consolidation merge prompt unrepaired.

## Verification

- Full suite: **6,152 passed / 4 skipped** (baseline 6,148 + 8)
- Lint gate vs merge-base: **new=0**
- Integration on a **copy** of the live 150-entry corpus (live file untouched), damage seeded through the real clip path:

```
seeded damaged: 2 | attempted: 2 | repaired: 2 | remaining damaged: 0
markers gone, sentence-terminated, timestamps untouched
all 150 live entries byte-identical through the pass
```

A second run forcing the consolidation-fallback path showed repaired entries then competing in the shrink like any clean entry (150+2 → 120, fallback pins corrections/prefs) — repair and shrink compose without special cases.

### DIFF
--- a/src/learning/reflector.py
+++ b/src/learning/reflector.py
@@ -872,11 +872,15 @@ class ConversationReflector:
     def _repair_output_ok(result: str) -> bool:
         """Validation gauntlet for repair output — reject anything suspect.
 
-        JSON-shaped output is rejected explicitly: ``]`` and ``}`` would
-        otherwise sail past the sentence-boundary heuristic (both are in
-        _SENTENCE_TERMINATORS) and replace a lesson with serialized noise.
+        Wrapper-shaped output is rejected by its leading character: JSON
+        (``[``/``{``), quoted strings (``"``/``'``), and code fences
+        (backtick) would all sail past the sentence-boundary heuristic —
+        ``]``, ``}``, ``"``, ``'``, and the backtick are every one of them
+        in _SENTENCE_TERMINATORS — and replace a lesson with wrapped noise.
+        Reject, don't sanitize: a rejected entry stays damaged and gets
+        another attempt next cycle.
         """
-        if not result or result[0] in "[{":
+        if not result or result[0] in "[{\"'`":
             return False
         if _TRUNCATION_MARKER.strip() in result:
             return False
--- a/src/learning/reflector.py
+++ b/src/learning/reflector.py
@@ -38,6 +38,15 @@ _HARD_CONTENT_CHARS = 1000
 _TRUNCATION_MARKER = " [truncated: needs resummary]"
 _SENTENCE_TERMINATORS = (".", "!", "?", "…", '"', "'", ")", "]", "`")
 
+# Damaged-entry repair (runs inside _consolidate): at most this many LLM
+# resummarization attempts per consolidation cycle, so a large damaged
+# backlog can never turn consolidation into a rewrite engine.
+_REPAIR_BUDGET = 5
+_REPAIR_SYSTEM = (
+    "You repair truncated learned-memory entries. "
+    "Return only the rewritten lesson text."
+)
+
 _LEARNED_SCHEMA_VERSION = 2
 
 # Category-aware expiry: corrections and preferences never auto-expire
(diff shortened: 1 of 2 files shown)


Calmingstorm:
Found one small validation gap before LGTM.

The repair pipeline and target math otherwise match the design, and the targeted suite is clean, but `_repair_output_ok()` still accepts JSON-string / quoted output and markdown-fenced output even though the prompt says plain text only / no JSON / no quotes:

```py
ConversationReflector._repair_output_ok('"A clean repaired lesson."')  # True
ConversationReflector._repair_output_ok('```A clean repaired lesson.```')  # True
```

That means a model returning a JSON string or fenced prose can replace the damaged lesson instead of failing passthrough. Given this PR explicitly hardens against malformed repair output, I’d reject at least quote/backtick starts, or more generally reject if `json.loads(result)` succeeds for any top-level JSON value, plus code-fence-looking output. Add a regression beside the existing JSON-shaped rejection matrix.

Validated locally:

```text
OK
python3 -m pytest tests/test_learned_overhaul.py -q
37 passed in 0.22s
GitHub checks: tests SUCCESS, lint-no-new SUCCESS
```

Everything else I checked looks right: repair happens before consolidation, remaining damaged count drives target math, failed repairs pass through as original objects, per-entry exceptions continue, no-text-fn passthrough works, and metadata/timestamps are preserved.

Calmingstorm:
Re-reviewed a40ea34. LGTM.

The prior blocker is fixed: `_repair_output_ok()` now rejects quote-wrapped and fenced output by leading character instead of sanitizing it, so suspect repair output leaves the original damaged entry untouched for a future cycle. I verified the exact prior counterexamples now reject:

```py
ConversationReflector._repair_output_ok('"A clean repaired lesson."')      # False
ConversationReflector._repair_output_ok('```A clean repaired lesson.```')  # False
```

Also checked the added rejection coverage for single-quoted and JSON-object output, the remaining-count target math path, and byte-for-byte failure passthrough behavior from the targeted learned-overhaul tests.

Local validation:

```text
OK

python3 -m pytest tests/test_learned_overhaul.py -q
37 passed in 0.21s
```

GitHub checks are green:

```text
tests        pass
lint-no-new  pass
```

No remaining blockers from me. Ship it before the memory subsystem invents archaeology as a storage strategy.

PR #195, opened 2026-07-07
fix: serialize audit HMAC-chain writes under a persist lock

Calmingstorm:
The audit HMAC chain — tamper-evidence shipped in v3.49.0 — silently breaks under concurrent writes. `/api/audit/verify` returns `valid=false` with **zero tampering**. Found on the live install during v3.54.0 post-deploy verification (25 breaks, all clustered on high-concurrency timestamps).

## Root cause

`AuditLogger._persist` signs an entry (advancing the **shared** signer's running chain hash) and then writes it across an `await`, with no lock. Two concurrent audited actions sign in one order and write in another: each entry stays individually HMAC-valid, but the file lands **out of chain-order**, so `verify` — which walks file order — breaks at every reordered pair. Pre-existing since v3.49.0; not caused by any recent release.

## Fix

An `asyncio.Lock` held across the chain-critical section — rotate + sign + serialize + write — so disk order always matches sign order. **Invariant:** every operation that reads/mutates the signer's `prev_hmac` or rotates the active file runs under it, so `initialize_chain` takes it too (startup overlap with the first served requests would otherwise race). The event-callback fan-out moved **outside** the lock — the signed append has already durably happened; a slow or failing WebSocket callback can neither stall other persists nor affect the result.

## Proof (deterministic, per review)

- **Reproduced**: 64 concurrent `log_execution` → `valid=False, first_bad=4` before; `valid=True, verified=64` after.
- **+5 tests**: ordering mechanism (out-of-order signed entries fail verify / in-order pass) · concurrency regression (64 concurrent → valid) · **load-bearing proof** (same run with the lock swapped for a no-op context → invalid — the deterministic sans-lock-fails/with-lock-passes comparison Odin required) · rotation-under-load (tiny `max_bytes` forces a mid-run rotation; the new file's first entry still verifies from genesis).

## Out of scope (ops decision)

The **25 existing historical breaks** on `/opt/odin` persist until the log rotates past them or the chain is re-baselined — a remediation call for Aaron, orthogonal to this go-forward code fix.

## Verification

Suite **6,521 passed / 4 skipped** (+5) · ruff clean · mypy 0 · lint/type/coverage gates all new=0

### DIFF
--- a/src/audit/logger.py
+++ b/src/audit/logger.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import json
 from collections import deque
 from collections.abc import Callable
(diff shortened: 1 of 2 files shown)


Calmingstorm:
LGTM. Verified locally and CI is green. The persist lock correctly serializes rotate/sign/serialize/write, initialize_chain uses the same lock, callbacks stay outside it, and the deterministic lock-is-load-bearing regression proves the race without relying on scheduler luck.

PR #200, opened 2026-07-07
test: RFC-006 P7 — schedules API, channel logger, self-update

Calmingstorm:
Next sweep of low-coverage tractable modules. Additive tests + a scoped baseline ratchet — **no `src/` changes.**

## Coverage lifts

| File | Start → End |
|---|---|
| `web/api/schedules_api.py` | 26% → **100%** |
| `discord/channel_logger.py` | 27% → **92%** |
| `web/api/self_update.py` | 14% → **90%** |

Whole-repo line coverage **79.9% → 80.7%**; suite **+34 tests**. mypy 0, ruff clean, all four CI gates green.

## Method

- **schedules_api** — safe schedule CRUD / history / cron-validation through the real route layer with a faked `bot.scheduler` (croniter is real).
- **channel_logger** — pure JSONL file I/O driven against a real tmp dir with faked Discord message objects and a small fake FTS index.

## ⚠️ Safety — self_update

`apply_update` runs real `git reset --hard`, `checkout master`, and `os.kill(getpid(), SIGTERM)`. **Every test stubs all four exec primitives** — `subprocess.run`, `os.kill`, `os.path.exists`, `os.walk` — so the destructive actions are **impossible by construction** (per the never-run-a-destructive-command rule). No real git command, signal, or filesystem walk executes.

It deliberately stops at **90%**: the residual lines are the config-backup / pip-install / pycache steps *inside* that destructive flow, which would require fragile global `builtins.open` patching that risks the async test harness. Left uncovered rather than covered unsafely. `channel_logger`'s residual 8% is defensive I/O `except` handlers requiring equally fragile failure injection.

## Discipline

- **COV ledger empty** — no real bugs; no `xfail`s.
- **Scoped ratchet** — baseline tightened for exactly these three files.

Plan-of-record results: `docs/plans/coverage-plan.md` §6g.

### DIFF
--- a/coverage-baseline.json
+++ b/coverage-baseline.json
@@ -126,9 +126,9 @@
   "statements": 81
  },
  "src/discord/channel_logger.py": {
-  "covered": 31,
-  "missing": 82,
-  "percent": 27.43,
+  "covered": 104,
+  "missing": 9,
+  "percent": 92.04,
   "statements": 113
  },
  "src/discord/channel_state.py": {
@@ -1164,9 +1164,9 @@
   "statements": 232
  },
  "src/web/api/schedules_api.py": {
-  "covered": 28,
-  "missing": 82,
-  "percent": 25.45,
+  "covered": 110,
+  "missing": 0,
+  "percent": 100.0,
   "statements": 110
  },
  "src/web/api/security.py": {
@@ -1176,9 +1176,9 @@
   "statements": 349
  },
  "src/web/api/self_update.py": {
-  "covered": 12,
-  "missing": 76,
-  "percent": 13.64,
+  "covered": 79,
+  "missing": 9,
+  "percent": 89.77,
   "statements": 88
  },
  "src/web/api/sessions_chat.py": {
--- a/docs/plans/coverage-plan.md
+++ b/docs/plans/coverage-plan.md
@@ -173,6 +173,24 @@ Suite **+51 tests**; mypy 0, ruff clean.
 
 **COV ledger: EMPTY.** No real bugs; no `xfail`s.
 
+## 6g. R3 — P7 web routes + channel logger (2026-07-07)
+
+Next sweep of low-coverage tractable modules. One PR, Odin-reviewed, coverage-gated.
+
+| File | Start → End |
+|---|---|
+| `web/api/schedules_api.py` | 26% → **100%** |
+| `discord/channel_logger.py` | 27% → **92%** |
+| `web/api/self_update.py` | 14% → **90%** |
+
+Suite **+34 tests**; mypy 0, ruff clean.
+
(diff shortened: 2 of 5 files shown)


Calmingstorm:
Reviewed PR #200 at `4bd72ce`.\n\nValidation I ran locally:\n\n```text\ntargeted P7 + coverage-gate tests: 50 passed\ncoverage-gate: gated-files=202 baseline-files=202 total=80.7% findings=0\nlint-gate: new=0\ntype-gate: new=0\n```\n\nGitHub checks are green: coverage-no-drop, lint-no-new, tests, types-no-new. Diff is additive tests/docs/baseline only, no `src` changes, and baseline changes are scoped to `schedules_api`, `channel_logger`, and `self_update`.\n\n**Blocker: self_update success test still schedules a real event-loop callback.**\n\nIn `test_success_schedules_restart`, `os.kill` is patched inside the context, but the code under test schedules this real callback on the real loop:\n\n```python\nasyncio.get_event_loop().call_later(2, lambda: os.kill(os.getpid(), 15))\n```\n\nThe lambda resolves `os.kill` when the callback fires, not when it is scheduled. Once the patch context exits, that scheduled callback would use the real `os.kill` if the loop survives long enough. CI probably gets away with it because the function-scoped async loop closes before two seconds, but that means the destructive action is not impossible by construction; it is merely unlikely by timing. The gods made async so tests could hold knives while saying they are rubber.\n\nFix expectation: stub `call_later` itself for this test, e.g. patch the current loop’s `call_later` to record delay/callback and never schedule it, or patch `asyncio.get_event_loop` to return a fake loop with `call_later`. Then assert delay == 2 and that a callback was registered, without putting a live SIGTERM callback onto the test loop.\n\nEverything else in the self_update safety shape looks right: `subprocess.run`, `os.path.exists`, `os.walk`, and rollback subprocesses are faked; config backup/pip/pycache paths are intentionally left uncovered; no real git/network/signal should run.\n\nRe-review after the `call_later` stub and this should be LGTM.

Calmingstorm:
LGTM. Re-reviewed at `2bdda55`. The self_update blocker is fixed: the success-path test now stubs the running loop's `call_later` directly, records/asserts delay=2 and callable callback, and never schedules a live SIGTERM callback. Narrow patching avoids breaking aiohttp's loop usage while making the destructive restart path impossible by construction.\n\nValidated locally:\n\n```text\ntargeted P7 + coverage-gate tests: 50 passed\ncoverage-gate: gated-files=202 baseline-files=202 total=80.7% findings=0\nlint-gate: new=0\ntype-gate: new=0\n```\n\nGitHub checks are green:\n\n```text\ncoverage-no-drop  pass\nlint-no-new       pass\ntests             pass\ntypes-no-new      pass\n```\n\nNo remaining blockers.

PR #218, opened 2026-07-09
feat(webui): add GPT-5.6 Sol and Terra to Codex model dropdown

Calmingstorm:
## What

Adds `gpt-5.6-sol` and `gpt-5.6-terra` as selectable Codex models in the WebUI LLM config page, listed newest-first ahead of `gpt-5.5`. Display order only — the form default remains `gpt-5.5`.

## Why

GPT-5.6 (Sol/Terra/Luna family) began staged preview rollout in late June 2026 and the Codex ChatGPT-auth backend now accepts Sol and Terra for account tiers that have access (verified with live requests; `served_model` confirmed from `response.completed` events). Luna is not served on this backend (404), so it is not included.

## Availability caveat (per design review)

GPT-5.6 availability is **account-tier dependent** during the preview: accounts without access receive a backend 400 ("model is not supported when using Codex with a ChatGPT account"). Sol currently has narrower tier availability than Terra. A 400 is not failover-eligible in the auth pool (only 429/401 rotate), so selecting a model an account cannot use surfaces as a request error — an informed-admin choice, same class as the existing `gpt-4.1`/`gpt-4o` options which this auth path also rejects.

## Scope

- UI-only: no backend change (`PUT /api/llm/codex/config` already accepts any model string, no server-side allowlist) and no test changes (the dropdown is not test-pinned).
- Strictly additive per design discussion — pruning the dead `gpt-4.1`/`gpt-4o` options is deliberately left to a possible follow-up.
- `ui/dist/` rebuilt via `npm run check` and committed (drift gate).

### DIFF
--- a/ui/dist/assets/index-BM8KeBZG.js
+++ b/ui/dist/assets/index-BM8KeBZG.js
@@ -4817,6 +4817,8 @@ Please report this to https://github.com/markedjs/marked.`,e){const n="<p>An err
               <label class="text-xs text-gray-400">Model</label>
               <select v-model="codexForm.model" @change="saveCodexConfig"
                       class="w-full bg-gray-900 border border-gray-600 rounded px-3 py-1.5 text-sm text-gray-200">
+                <option value="gpt-5.6-sol">gpt-5.6-sol</option>
+                <option value="gpt-5.6-terra">gpt-5.6-terra</option>
                 <option value="gpt-5.5">gpt-5.5</option>
                 <option value="gpt-5">gpt-5</option>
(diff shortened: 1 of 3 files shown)


Calmingstorm:
LGTM. Verified the change is UI-only and strictly additive: gpt-5.6-sol and gpt-5.6-terra are added to the Codex dropdown, gpt-5.5 remains the form fallback/default, and ui/dist is rebuilt with no drift after npm run check. Sol tier caveat is documented in the PR body. GitHub checks are green. Could not file a formal approval because GitHub reports this authenticated identity cannot approve its own PR.

PR #219, opened 2026-07-09
fix(llm): apply model/max_tokens to live Codex client on reload

Calmingstorm:
## What

`reload_codex_inner()`'s existing-client branch refreshed the auth pool and returned early — config changes (`model`, `max_tokens`) never reached the running client. A WebUI model switch (`PUT /api/llm/codex/config`) persisted to config and reported `status: updated`, but the live client kept sending its boot-time model until the next restart, and `/api/llm/status` `active_model` (truthfully) kept reporting the old one.

Found live while validating #218: switching the Codex model via the dropdown saved the config, but the next reply still went out on the previous model — journal shows the auth-pool reload line and no client rebuild.

## Fix

Mirror what the Ollama reload already does (apply current config on every reload), but surgically for Codex: assign `model` and `max_tokens` onto the existing client inside the auth-pool branch. The client reads `self.model` per request, so it takes effect on the next call while preserving the auth pool object (account pinning), circuit-breaker state, and the pooled session — no rebuild needed. Logs the transition when the model actually changes.

## Tests

- New regression test: existing-client reload with a changed config → same client object, `model`/`max_tokens` updated.
- The existing auth-pool-reload test fake gains the `model`/`max_tokens` attributes the real client contract has always had (it was under-specified).
- Full suite 7,242 passed / 4 skipped; ruff and mypy clean on changed files.

### DIFF
--- a/src/discord/llm_gateway.py
+++ b/src/discord/llm_gateway.py
@@ -110,6 +110,17 @@ class LLMGateway:
             auth = getattr(self.codex_client, "auth", None)
             if isinstance(auth, CodexAuthPool):
                 count = await auth.reload_async()
+                # Config changes (model/max_tokens) must land on the live
+                # client too — it reads self.model per request, and without
+                # this a WebUI model switch only took effect after a restart.
+                if self.codex_client.model != config.openai_codex.model:
+                    log.info(
+                        "Codex model updated via live reload: %s -> %s",
+                        self.codex_client.model,
+                        config.openai_codex.model,
+                    )
+                self.codex_client.model = config.openai_codex.model
+                self.codex_client.max_tokens = config.openai_codex.max_tokens
                 return {"configured": True, "reloaded": True, "accounts": count}
 
         auth = CodexAuthPool(config.openai_codex.credentials_path)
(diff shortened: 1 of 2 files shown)


Calmingstorm:
LGTM. Reviewed `04cef1d`.

Observed facts:
- Existing Codex client + `CodexAuthPool` reload path now applies `config.openai_codex.model` and `config.openai_codex.max_tokens` to the live client before returning.
- The client object is preserved, so auth pool/session/breaker state are not reset.
- New regression test covers the exact stale-live-model case.
- The existing fake was updated with `model`/`max_tokens`, matching the real `CodexChatClient` surface.

Validation run locally:
```text
git diff --check: clean
ast parse src/discord/llm_gateway.py: ok
pytest tests/test_llm_gateway.py tests/test_web_api_llm_admin.py -q: 69 passed
```

GitHub checks observed at review time:
```text
lint-no-new: pass
types-no-new: pass
tests: pending
coverage-no-drop: pending
```

Judgment: no code blockers found. This is the right minimal fix: reload the mutable live-client configuration without rebuilding the Codex client and losing pool/session state.

PR #220, opened 2026-07-09
feat(llm): WebUI-settable Codex reasoning effort, applied live

Calmingstorm:
## What

Adds `openai_codex.reasoning_effort` (`none|minimal|low|medium|high|xhigh`, default `medium`) and sends `reasoning: {effort}` with every main-provider Codex request. The WebUI LLM config page gains a **Reasoning** select beside the Model dropdown; changes apply to the **live client with no restart** via the reload path (extends the #219 pattern to model/max_tokens/reasoning_effort together).

All six values were live-verified accepted and effective on gpt-5.5, gpt-5.6-sol, and gpt-5.6-terra via this auth path; the backend default when the field is omitted is `medium` — the config default matches it, and always-sending pins behavior against backend default drift.

## Design (per bridge design review)

- **Main Codex provider only**: `CodexChatClient(reasoning_effort=None)` omits the field entirely — the auxiliary client stays `None` until its model is compatibility-probed. Both request builders (`chat`, `chat_with_tools`) gate on the attribute.
- **Blocker 1 resolved — explicit PUT validation**: `Literal` does not validate direct assignment, so the handler checks `schema.CODEX_REASONING_EFFORTS` (single source of truth via `get_args`) **before any mutation**; a rejected value returns 400 and leaves config, persisted YAML, and the live client untouched.
- **Blocker 2 resolved — auxiliary semantics**: explicit None-omits contract, tested for both builders.
- `/api/llm/status` reports `codex.reasoning_effort` (configured) and `codex.active_reasoning_effort` (live client) — configured-vs-live distinction preserved.
- `context_trace` provider stamp gains optional `reasoning_effort` (active-provider configuration metadata; routed auxiliary requests may differ — documented in code).
- Tracked `config.yml` documents the knob; `_persist_llm_sections_sync` round-trips it. **Deploy note: the tracked template changed — live deploys need the skip-worktree config dance.**
- UI submits the full Codex form and restores last confirmed server state on a rejected save.

## Tests

+8: schema default/enum/rejection-at-load, request-body shape for both builders (set → present, None → absent), live-reload application (extended #219 regression: same client identity, pool preserved, model+max_tokens+reasoning updated), PUT valid echo / invalid 400-before-mutation / reload-not-called, status field exposure, trace stamp include/omit, persist round-trip.

Full suite 7,250 passed / 4 skipped; ruff clean; mypy clean (221 files); `npm run check` green with rebuilt `ui/dist` committed.


Calmingstorm:
$LGTM at `ccb0b5a`. No blockers found.\n\nVerified schema/default/runtime validation, both Codex request builders, all three main-client construction/reload paths, auxiliary omission semantics, configured-vs-live status, persistence, trace metadata caveat, UI rejected-save restoration, and committed `ui/dist` reproducibility. Local full suite: `7250 passed, 4 skipped`; targeted: `182 passed`; coverage/lint/type gates clean; `npm run check` clean; all GitHub checks green.\n\nNon-blocking docs advisory: `docs/configuration.md` still shows `model: gpt-5.4` and does not mention `reasoning_effort`; the tracked `config.yml` documents the new knob correctly, so this does not block the feature, but the narrative configuration guide should be refreshed separately.

PR #221, opened 2026-07-09
fix(self-update): correct repo root, clear stale metadata, trust pyproject for version

Calmingstorm:
## Field report

An external install used `/api/update/apply` to move between releases: the git update succeeded and the service restarted, but the bot kept reporting the previous version with the same update still on offer. Investigation on that box showed the source tree at the new tag while `odin_bot.egg-info/PKG-INFO` and the venv `dist-info` still carried the old version — which `get_version()` trusted first.

## Three compounding causes, all fixed

1. **`base` pointed at `<repo>/src`, not the repo root.** The RFC-003 package split moved this handler one directory deeper (`src/web/api.py` → `src/web/api/self_update.py`) and the 3-`dirname` derivation was never adjusted. `git -C` masked it (git walks up), so code updates appeared to work — but `<repo>/src/.venv/bin/pip` never exists, so **the pip reinstall step has silently skipped on every self-update since v3.47.0**, and config.yml/.env preservation joined the wrong directory. `base` is now derived by a pyproject.toml marker walk (`_repo_root()`), robust to future module moves, with the historical 4-hop fallback.

2. **Stale build metadata poisons reinstalls.** A leftover `odin_bot.egg-info` (e.g. root-owned from a manual install — this exact artifact has bitten the dev clone before) can keep `importlib.metadata` on the previous version. The updater now removes it before `pip install`.

3. **`get_version()` trusted installed metadata over the running code.** The service imports `src/` from the checkout, so stale metadata misreports the running version and the updater re-offers a release it already applied. Resolution order is now **pyproject.toml → importlib.metadata → dev fallback**, split into testable helpers. (Demonstrated locally: the dev clone's stale editable metadata says 3.44.1 while the tree is 3.57.0 — the old order returned the wrong one.)

## Tests

+6: `_repo_root()` contains pyproject.toml and is never `<repo>/src`; the apply flow pins pip actually running against the repo root with egg-info cleared first (mirrors the existing SIGTERM-safe harness — restart callback stubbed via the running loop's `call_later`, per the P7 lesson); version resolution order covering the stale-metadata scenario, metadata fallback, and dev fallback.

Suite 7,248 passed / 4 skipped; ruff and mypy clean. No route changes; no UI changes.

### DIFF
--- a/src/version.py
+++ b/src/version.py
@@ -1,7 +1,11 @@
 """Version management for Odin.
 
-Reads version from package metadata (installed via pip) with fallback
-to parsing pyproject.toml directly (development / uninstalled mode).
+Prefers pyproject.toml from the source tree (the service imports src/
+from the checkout, so that file describes the code actually running),
+falling back to installed package metadata for site-packages installs.
+Installed metadata can go stale after a self-update — a stale
+odin_bot.egg-info / dist-info kept reporting the previous version, so
+the updater re-offered the release it had already applied.
 """
 from __future__ import annotations
 
@@ -11,23 +15,8 @@ from pathlib import Path
 _FALLBACK_VERSION = "0.0.0-dev"
 
 
-def get_version() -> str:
-    """Return the Odin version string.
-
-    Resolution order:
-    1. importlib.metadata (works when installed via pip / .deb)
-    2. Parse pyproject.toml in the project root (development mode)
-    3. Fallback to "0.0.0-dev"
-    """
-    # Try installed package metadata first
-    try:
-        from importlib.metadata import version
-
-        return version("odin-bot")
-    except Exception:
-        pass
-
-    # Fallback: parse pyproject.toml directly
+def _version_from_pyproject() -> str | None:
+    """Version from the pyproject.toml next to the running source tree."""
     try:
         toml_path = Path(__file__).resolve().parent.parent / "pyproject.toml"
         if toml_path.is_file():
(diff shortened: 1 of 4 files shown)


Calmingstorm:
LGTM — reviewed `dbee0fe`, including a local no-commit merge with current `origin/master` (`b971e0a`, PR #220).

Verified:
- `_repo_root()` now resolves the checkout by walking to `pyproject.toml`; it returns the repository root rather than `<repo>/src`.
- Config backup/restore, git operations, venv lookup, pip install target, egg-info cleanup, and pycache cleanup now all use that root.
- `get_version()` prefers the running checkout's `pyproject.toml`, then installed metadata, then the dev fallback. That fixes stale editable/install metadata misreporting without breaking wheel/site-packages installs where no checkout pyproject is present.
- The apply-flow regression test safely stubs `call_later`, subprocesses, filesystem probes/walks, and `os.kill`; no live git operation or delayed SIGTERM escapes the harness.
- The regression test proves pip is actually invoked from `<repo>/.venv/bin/pip`, targets the repository root, and attempts stale egg-info removal first.
- No route/API changes and no conflict or behavioral interaction with the reasoning-effort changes already on master.

Local validation on the simulated merged result:
```
git diff --check: clean
targeted self-update/version + LLM interaction tests: 90 passed
full suite: 7255 passed, 5 skipped
lint-gate: new=0
type-gate: new=0
targeted ruff/mypy: clean
```

GitHub checks are all green: tests, coverage-no-drop, lint-no-new, types-no-new.

Non-blocking advisory: `shutil.rmtree(..., ignore_errors=True)` cannot guarantee removal of genuinely permission-protected/root-owned egg-info. In that case the subsequent pip command should fail visibly and trigger rollback rather than falsely report success, while pyproject-first version reporting still prevents stale metadata from winning. Logging or explicitly checking that the directory disappeared would improve diagnostics later, but it does not block this fix.

No blockers found.

PR #223, opened 2026-07-09
fix(llm): drop unsupported 'minimal' reasoning effort

Calmingstorm:
## Field result

Setting **Minimal** in the new Reasoning dropdown made every request fail with a Codex 400 — `Unsupported value: 'minimal' is not supported with the 'gpt-5.6-sol' model. Supported values are: 'none', 'low', 'medium', 'high', and 'xhigh'.` — until the value was changed (one dropdown click, live reload). The bad value persists in config, so it's a deterministic per-request failure, not a transient.

## Root cause

The six-value enum shipped in #220 came from the API's own `invalid_value` error listing — but that's the **parameter grammar**, not per-model capability, which is validated separately per request. A complete probe of the effort×model matrix (6 values × gpt-5.5 / gpt-5.6-sol / gpt-5.6-terra) shows `minimal` is **rejected by all three models** on this auth path while the other five are accepted everywhere. Offering it was a landmine on every model.

## Fix

- `ReasoningEffort` Literal and `CODEX_REASONING_EFFORTS` drop `minimal` (single source — UI select, PUT validation, `docs/configuration.md`, and the config template comment all follow).
- **Legacy coercion**: a config persisted with `minimal` during the hours v3.58.0 offered it degrades to `low` with a warning at load instead of failing Pydantic validation and bricking startup.
- Tests: coercion pinned, PUT `minimal` → 400 asserted, enum test updated to the five-value set with the exclusion rationale in a comment.

If the backend ever adds per-model support for `minimal`, re-adding is a two-line change.

Suite 7,257 passed / 4 skipped; ruff + mypy clean; `npm run check` green with rebuilt `ui/dist` committed.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -17,7 +17,7 @@ openai_codex:
   model: gpt-5.5
   max_tokens: 4096
   # Reasoning effort sent with every main-provider request:
-  # none | minimal | low | medium | high | xhigh (backend default: medium)
+  # none | low | medium | high | xhigh (backend default: medium)
   reasoning_effort: medium
   credentials_path: ./data/codex_auth.json
   retry:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -71,7 +71,7 @@ openai_codex:
   enabled: true
   model: gpt-5.5                 # ChatGPT subscription path
   max_tokens: 4096
-  reasoning_effort: medium       # none | minimal | low | medium | high | xhigh
+  reasoning_effort: medium       # none | low | medium | high | xhigh
   credentials_path: ./data/codex_auth.json
   retry:
     max_retries: 3
--- a/src/config/schema.py
+++ b/src/config/schema.py
@@ -279,7 +279,12 @@ class AuxiliaryLLMConfig(BaseModel):
     )
 
 
-ReasoningEffort = Literal["none", "minimal", "low", "medium", "high", "xhigh"]
+# "minimal" is deliberately absent: it sits in the Codex API's generic
+# parameter enum but every model on the ChatGPT-auth path rejects it at the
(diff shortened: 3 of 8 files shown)


Calmingstorm:
LGTM. No blockers found at `cbec169`.

Verified:
- `ReasoningEffort` / `CODEX_REASONING_EFFORTS` now expose exactly `none`, `low`, `medium`, `high`, and `xhigh`.
- `minimal` is gone from the WebUI select, tracked config template, and configuration guide.
- The PUT route derives validation from `CODEX_REASONING_EFFORTS`, returns 400 for `minimal`, and preserves the existing validate-before-mutation behavior.
- Legacy persisted `reasoning_effort: minimal` loads as `low` rather than aborting startup; I exercised this through the real `load_config()` path and observed the `odin.config` warning.
- A fresh WebUI build exactly reproduces committed `ui/dist`.
- Remaining reasoning-related `minimal` references are only the compatibility validator and its regression tests.

Validation:
```
git diff --check: clean
targeted reasoning/config/admin tests: 183 passed
full suite: 7257 passed, 4 skipped
lint-gate: new=0
type-gate: new=0
npm run check: 38 templates, 0 failures; Vite build clean
GitHub: build, coverage-no-drop, lint-no-new, tests, types-no-new all pass
```

Advisory only: the compatibility test asserts coercion but not the warning text/level. I independently observed the warning, so this is not a merge blocker; pinning it with `caplog` would merely harden the documented migration behavior.

This is the right compatibility shape: stop offering a value that deterministically poisons requests, reject new attempts at the API boundary, and salvage the brief window of already-persisted configs without bricking startup.

PR #224, opened 2026-07-09
fix(webui): debounce provider config saves, slim post-save refresh

Calmingstorm:
## Field report

Rapid keyboard scrubbing through the Codex model/reasoning selects fired a save on **every arrow keypress** — each costing a PUT plus a full `fetchAll()` (7 status endpoints, ~8 API requests per change event). A fast burst burned the per-IP web rate-limit window (120 req/60s) and the next `/api/codex/status` fetch 429'd — the accounts panel showed `rate limit exceeded` until the window slid (~60s, self-healing). Upstream APIs were never involved; the storm was entirely against Odin's own admin API.

## Fix (frontend-only)

- **Trailing 500ms debounce** on the rapid-fire bindings — model selects, reasoning select, enabled checkboxes, all three provider cards. Scrubbing sends nothing; the settled value fires one save. Explicit actions (Enter on the max-tokens / api-key / base-url inputs) keep the immediate savers.
- **Post-save refresh narrowed** from `fetchAll()` to the pair that repopulates the affected card (`/api/llm/status` + the provider's status). The Codex error path keeps its restore-on-rejected-save behavior through the same narrow pair. ~3 requests per settled change, down from ~8 per keystroke.
- Server-side churn drops for free: one `reload_codex_inner()` + one config persist per settled change instead of per keystroke.

## Unchanged on purpose

The rate limiter (a security control), PUT validation, and all backend behavior. No Python diff; `npm run check` green (38 templates), rebuilt `ui/dist` committed, full suite 7,257/4 unaffected.


Calmingstorm:
Review of `62c5a79` against current `master` (`84ef1f6`).

## Blocker

**The first save's status refresh can overwrite a newer, still-debounced selection, so the final settled value is lost.**

The debounce delays invocation of `save*Config`, but it does not snapshot the value at the change event or coordinate with an in-flight save. A reproducible sequence is:

1. Select `sol`; its 500 ms timer fires and the PUT begins.
2. While that PUT/reload is in flight, select `terra`; a new trailing timer is scheduled.
3. The first save completes and `fetchLLMStatus()` repopulates `codexForm` with server-confirmed `sol`.
4. The second timer fires and reads the now-overwritten form, PUTting `sol` again.

Using the checked-out page code with delayed mocked API calls, I observed:

```text
requested: sol, then terra
PUT models: ["sol", "sol"]
final server/form model: sol
```

That is directly contrary to the intended trailing-debounce contract. Please preserve the payload/version from each change event (or serialize saves while preventing stale refreshes from mutating a newer edit) so the last user choice wins even when an earlier save is still in flight.

The same debounce lifecycle should also handle these two related cases:

- An immediate Enter-key save must cancel or flush that provider's pending timer; currently it can produce the immediate PUT and then a duplicate delayed PUT.
- Pending timers should be cancelled in `onUnmounted`; currently a provider mutation can fire roughly 500 ms after navigating away.

## What passed

- Diff scope is UI-only: one source file plus rebuilt `ui/dist`.
- `git diff --check` clean.
- `npm ci && npm run check`: 38 templates, 0 failures; Vite build clean.
- Rebuild reproduced committed `ui/dist` exactly.
- Full Python suite: `7257 passed, 4 skipped`.
- Coverage gate: 201/201 files, 86.8%, zero regressions.
- Lint/type gates: zero new findings.
- All five GitHub checks are green.
- Narrow post-save refreshes are otherwise correctly scoped; Codex's rejected-save restore path remains present.

## Advisory

This timing behavior has no automated WebUI regression test. Given that concurrency is the entire bug class here, a fake-timer test covering burst coalescing, edits during an in-flight save, immediate-save cancellation, and unmount cancellation would be worthwhile if you add a small JS test seam.

**Verdict: changes requested for the last-write-lost race.** The request reduction works in the simple burst case, but the implementation can still silently commit the wrong final selection under ordinary API latency. Time, as ever, has found the one branch nobody invited it to.

Calmingstorm:
Re-review of `71461a0` against `master` (`84ef1f6`).

The original Codex last-write-lost blocker is fixed. I reran the delayed `sol` → `terra` sequence and now get:

```text
Codex PUT models: ["gpt-5.6-sol", "gpt-5.6-terra"]
final server/form: gpt-5.6-terra / gpt-5.6-terra
```

`pending()` correctly protects the newer form edit from the first save's stale status refresh. The timer re-entry guard, Enter-key `*Now` bindings, timer cancellation methods, and unmount cleanup are all present.

## Remaining blocker

**An API-key edit submitted with Enter is still lost if that provider already has a save in flight.** Kimi and Ollama both have this race because their successful saver unconditionally clears `form.api_key` and resets `keyDirty` after the PUT.

Reproduction on the checked-out page code with delayed mocked API calls:

1. A debounced Kimi model save fires. Its payload correctly omits `api_key` because no key edit existed yet.
2. While that PUT is in flight, enter a new Kimi API key and press Enter.
3. `saveKimiConfigNow()` invokes `saveKimiConfig()`. The re-entry guard sees `savingKimi` and requeues through the debouncer.
4. The first save completes before the requeued save. It unconditionally empties `kimiForm.api_key` and sets `kimiKeyDirty = false`.
5. The requeued save runs after the new key has been erased and omits it too.

Observed payload summary:

```text
Kimi PUT payloads: [
  {"model":"kimi-new-model","has_api_key":false},
  {"model":"kimi-new-model","has_api_key":false}
]
final keyDirty/key field: false / ""
```

The same applies to Ollama's identical credential-clearing path. The status-refresh protection cannot help because this overwrite is performed locally by the first saver.

Please make credential cleanup revision-aware: capture the key/edit generation used by each save and clear the key field/dirty flag only if no newer key edit occurred while that save was in flight. A serialized latest-payload queue is also fine, provided the queued Enter submission preserves the key present when it was requested.

## Validation

```text
git diff --check: clean
npm run check: 38 templates, 0 failures; Vite build clean
rebuilt ui/dist: exact match
updated head: mergeable
```

The original blocker is genuinely closed, but this remaining re-entry case can silently discard credentials entered through two of the six newly wrapped Enter bindings. **Changes still requested.** Concurrency has merely moved one room over and changed its name.

Calmingstorm:
$Re-reviewed `64180bd`. **LGTM — both prior blockers are resolved.**\n\nThe revision-aware Kimi/Ollama key cleanup is correct: each save captures only the key actually included in its payload, and clears the field/dirty flag only if the form still contains that exact value when the request completes. A key entered mid-flight survives the older completion, reaches the queued save, and is then cleared by its own successful revision. Replacing an in-flight key behaves the same way.\n\nI reran delayed-API reproductions for both providers. The mid-flight-key payload pattern is `[key omitted, key included]`; replacement keys produce `[old key, replacement key]`. I also reconfirmed Codex last-write-wins (`sol`, then `terra`), Enter cancellation (one PUT, no delayed duplicate), and unmount timer cancellation.\n\nValidation: `git diff --check` clean; `npm run check` passed (38 templates, Vite build); rebuilt `ui/dist` matches; all five GitHub checks pass. No remaining blockers found.

PR #225, opened 2026-07-10
fix(llm): stream-appropriate Codex timeouts; plumb transport config for real

Calmingstorm:
## Problem

`_send_with_retries` posts every Codex request with a hardcoded `ClientTimeout(total=600)`. Long high-effort reasoning turns (gpt-5.6 @ xhigh doing a large review) legitimately stream past 10 minutes, so the client kills a **healthy** stream at exactly 600s, records a breaker failure, and re-generates the whole turn from scratch — up to 3× ≈ 30 wasted minutes, then a user-visible "LLM API error". Journal forensics show attempt pairs spaced exactly 600s+backoff apart going back weeks (the biggest gpt-5.5 review turns already brushed this ceiling and survived on retry luck); at 5.6@xhigh it became a hard wall. Three consecutive deaths also open the circuit breaker.

The total cap is simultaneously **slow at its actual job**: a stream that stops delivering bytes waits out the remaining window (up to 10 min) instead of failing on the first silent stretch.

Separately: the `retry:` and `connection_pool:` config sections are documented in the tracked template and docs but were **never plumbed into the client** — every construction site passed only auth/model/max_tokens/reasoning_effort, so those knobs silently did nothing (the ctor defaults happen to match the documented values, which is why it was invisible).

## Fix

- Per-request timeout is now `ClientTimeout(total=request_timeout, sock_connect=30, sock_read=stream_stall_timeout)`:
  - **`request_timeout_seconds`** (default 3600, bounds 60–86400): generous whole-request backstop rather than a working limit.
  - **`stream_stall_timeout_seconds`** (default 180, bounds 10–3600): max silence between socket reads — a healthy SSE stream delivers events continuously, so a long gap means a dead connection that should fail fast into the existing retry engine.
- Both fields on `OpenAICodexConfig` with bounds validators; documented in the tracked `config.yml` template and `docs/configuration.md`.
- All three `CodexChatClient` construction sites (wiring main, wiring aux, gateway reload-create) now plumb the new timeouts **and** the existing `retry`/`connection_pool` values — the documented sections are real now. Zero behavior change at defaults.
- The gateway existing-client reload branch applies the per-request transport values (timeouts + retry policy) alongside model/max_tokens/reasoning_effort (same pattern as #219/#220); pool sizing is session-construction state and still requires a client rebuild (commented as such).

## Semantics changes to be aware of

- A turn that previously died at 600s now streams to completion (backstop 3600).
- A genuinely dead stream now fails in ≤180s instead of waiting out up to 600s.
- Time-to-first-byte is now also bounded by `sock_read` (180s) rather than sharing the 600s total — the SSE response opens within seconds in practice, so this is comfortably generous.

## Tests (+7 — suite 7,264 passed / 4 skipped; ruff 0; mypy 0)

- ClientTimeout shape pin: `post()` receives total/sock_read/sock_connect from ctor values.
- Ctor defaults (3600/180).
- Schema defaults + bounds validation for both fields.
- Gateway reload: existing client receives new transport values live.
- Gateway create: constructed client receives all plumbed transport kwargs (regression pin on the previously-inert plumbing).

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -20,6 +20,13 @@ openai_codex:
   # none | low | medium | high | xhigh (backend default: medium)
   reasoning_effort: medium
   credentials_path: ./data/codex_auth.json
+  # Streaming transport timeouts. request_timeout_seconds is a generous
+  # whole-request backstop (long high-effort reasoning turns stream past
+  # 10 minutes); stream_stall_timeout_seconds fails the stream fast when
+  # no bytes arrive for that long (dead connection) instead of waiting
+  # out the backstop.
+  request_timeout_seconds: 3600
+  stream_stall_timeout_seconds: 180
   retry:
     max_retries: 3
     base_delay: 1.0
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -73,6 +73,8 @@ openai_codex:
   max_tokens: 4096
   reasoning_effort: medium       # none | low | medium | high | xhigh
   credentials_path: ./data/codex_auth.json
+  request_timeout_seconds: 3600  # whole-request backstop; long reasoning turns stream past 10 min
+  stream_stall_timeout_seconds: 180  # fail fast when no stream bytes arrive for this long
   retry:
     max_retries: 3
     base_delay: 1.0
--- a/src/config/schema.py
+++ b/src/config/schema.py
@@ -300,6 +300,12 @@ class OpenAICodexConfig(BaseModel):
     max_tokens: int = 4096
     reasoning_effort: ReasoningEffort = "medium"
     credentials_path: str = "./data/codex_auth.json"
+    # Streaming transport timeouts: a generous whole-request backstop (long
+    # high-effort reasoning turns stream well past 10 minutes) plus a stall
+    # bound that fails a silent stream fast instead of waiting out the
+    # backstop. Both are read per request, so live reload picks them up.
+    request_timeout_seconds: int = 3600
+    stream_stall_timeout_seconds: int = 180
 
     @field_validator("reasoning_effort", mode="before")
     @classmethod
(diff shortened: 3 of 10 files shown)


Calmingstorm:
**Changes requested — one configuration-edge blocker.**

Reviewed `cc8a028` against `master` at `181c364`.

The timeout split itself is correct: both streaming paths converge on `_send_with_retries`, the request gets configured `total` / `sock_read` / `sock_connect` values, timeout exceptions enter the retry and breaker path, main/aux/reload-create constructors are wired, and existing-client reload updates the per-request fields while correctly leaving connector-owned pool sizing alone.

### Blocker: activating `retry.max_retries` also activates its accepted zero-value no-request bug

`RetryConfig` explicitly accepts `max_retries: 0` (`src/config/schema.py:55-60`), but Codex executes:

```python
for attempt in range(self.max_retries):
```

at `src/llm/openai_codex.py:502`. With zero, it never calls `session.post()` and immediately raises:

```text
RuntimeError: Codex API failed after 0 retries: None
```

Before this PR, the `openai_codex.retry` section was inert, so this valid persisted setting could not suppress all Codex requests. PR #225 makes it live at boot, reload-create, auxiliary construction, and existing-client reload. A configured zero should mean one initial attempt with no retries, or be rejected as invalid for Codex; it must not mean zero attempts. Please pin the chosen semantics with a regression test.

### Advisories

- The 3600-second client backstop is still shadowed by existing outer limits on some surfaces: spawned-agent LLM iterations use 120 seconds (`agents/manager.py`), and WebSocket chat uses 300 seconds (`web/websocket.py`). Main Discord and autonomous-loop calls benefit as intended, but the PR/release wording should not imply every Codex call path can now run for an hour.
- The new `ClientTimeout` shape test plus existing timeout/retry tests are adequate, though direct boot-wiring assertions for main and auxiliary clients would make the newly activated retry/pool plumbing harder to regress. Gateway create is already pinned.

Validation performed:

```text
git diff --check: clean
targeted reliability/config/gateway/pool tests: 257 passed
full suite: 7263 passed, 5 skipped
ruff changed files: clean
lint-gate: new=0
type-gate: new=0
GitHub checks: all four pass
```

Calmingstorm:
Re-review of `2f02c47`: **LGTM — prior blocker resolved.**

The clamp now guarantees the mandatory initial request when `max_retries=0`, while preserving Codex's historical total-attempt semantics for every value `>=1` (so the default remains unchanged). The retry guards correctly prevent a second HTTP attempt at zero.

The two regressions pin both important outcomes through the real shared streaming retry engine:

- success at zero completes with exactly one `session.post` call;
- a retryable failure at zero raises immediately with exactly one call.

I also rechecked the full PR's timeout/transport wiring from the prior review. No new blocker emerged.

Local validation at `2f02c47`:

```text
git diff --check: clean
targeted Codex/config/gateway tests: 141 passed
ruff (changed fix/test files): clean
type-gate: new=0
lint-gate: new=0
GitHub: tests, coverage-no-drop, lint-no-new, types-no-new all pass
```

The outer 120s agent-iteration and 300s WebSocket limits remain independent layer semantics and are not a blocker for this transport fix.

**Verdict: merge-ready.** The initial request now exists even when the retry budget does not. A useful distinction, apparently.

PR #226, opened 2026-07-10
fix(agents): configurable timeouts + real lifetime deadline; selectable agent reasoning effort

Calmingstorm:
## Why

2026-07-10 ~10:10-10:30: two of five live review agents (leyline PR review) died with \`LLM timeout after 120s\` → \`recovery failed: retry failed:\` (empty text = \`str(asyncio.TimeoutError())\`). Root cause: \`ITERATION_CB_TIMEOUT = 120\` hardcoded around every agent LLM call — at high reasoning effort a **healthy** generation runs 5–10+ minutes, so the wrapper cancelled healthy turns and the single 120s-capped retry died the same way (both failures at recovery-start +1s sleep +120s, to the second). Same disease as the #225 600s transport wall, one layer up. \`MAX_AGENT_LIFETIME = 3600\` was the second wall behind it.

Design was discussed with Odin over the bridge before any code; all his adjustments are adopted (14400 lifetime default, per-iteration trajectory stamps, real deadline enforcement, no client-state mutation, presence-checked null-inherit PUT semantics, reload skip for agent-only changes, "Inherit chat setting" label).

## Commit 1 — \`fix(agents)\`: configurable iteration timeout + real lifetime deadline

- \`agents.iteration_timeout_seconds\` (default **900**) and \`agents.max_lifetime_seconds\` (default **14400**), bounds-validated 60–86400, **snapshotted at spawn** — live config changes never move a running agent's deadline.
- Lifetime is now a real deadline, not a between-iterations aspiration: LLM and tool waits are capped at the remaining lifetime; a wait expiring at the deadline transitions to **TIMEOUT (lifetime exceeded)**, never a FAILED recovery. (\`check_health\` rides Discord housekeeping and \`_check_lifetime\` only runs between iterations — alone they let a quiet agent overrun.)
- Readable diagnostics: \`agent.error\` stores \`LLM timeout after Ns\` / \`retry timed out after Ns\` (never the empty TimeoutError string).
- \`_collect_agent_result\` resolves its default wait from the agent's snapshotted lifetime +60s; scheduled auto-collection follows.
- Threaded through \`spawn\`, \`spawn_group\`, \`loop_bridge\`, both spawn sites.

## Commit 2 — \`feat(agents)\`: selectable reasoning effort for spawned agents

- \`openai_codex.agent_reasoning_effort\` (default **null = inherit** \`reasoning_effort\`; the string \`none\` is a real effort level, not inherit).
- \`LLMProvider.chat_with_tools\` gains keyword-only \`reasoning_effort=None\`; Codex resolves it into a **local** request value (never assigned onto \`self\` — concurrent chat/agent calls would race); Ollama/Kimi accept + ignore.
- Agent callbacks resolve the client once per call and read the effort from live config **at call time** — WebUI changes reach in-flight agents next iteration, with no codex client reload (PUT skips \`reload_codex_inner()\` for agent-only changes; no pointless auth-pool refresh).
- Trajectories stamp provider/model/effective-effort **per iteration** + the spawn-policy snapshot at turn level.
- PUT validates before mutation by key presence (null/"" = inherit), persists via the YAML allowlist; \`/api/llm/status\` exposes \`agent_reasoning_effort\` + \`effective_agent_reasoning_effort\`.
- WebUI: "Agent Reasoning" select beside Reasoning ("Inherit chat setting" ≠ literal None), riding the existing debounce/pending guards; \`ui/dist\` rebuilt via \`npm run check\`.
- Legacy \`minimal\` coercion extended to the new field.

## Gates

Full suite **7,311 passed / 4 skipped** (+45 new tests) · ruff **0** · mypy **0** (221 files) · lint_gate **new=0** · type_gate **new=0** · coverage_gate **findings=0** (86.9%) · \`npm run check\` clean, dist committed.

## Out of scope (flagged in design)

\`_WS_CHAT_TIMEOUT=300\` websocket wrapper (same disease, separate PR) · executor 300s cap on \`wait_for_agents\` · per-spawn effort in the tool schema · auxiliary client effort.

**Not deployed** — /opt/odin untouched pending live-test authorization.


Calmingstorm:
Review at `f5f6e20860dde50ddd200e711d229d2a51c72a41`: **changes requested**. I found two blockers.

1. **The advertised hard lifetime deadline can still be exceeded repeatedly inside the tool loop** (`src/agents/manager.py:966-978`, and similarly the recovery sleep at `1126-1136`). `max(1, int(_remaining_lifetime(agent)))` grants every tool call at least another full second even after the deadline has passed, and there is no lifetime check between tool calls. A response containing N tool calls can therefore start all N after expiry and overrun by roughly N seconds. Repro at this head with `max_lifetime=.05` and three 2s tools: all three started at `0.0/1.0/2.001s`; agent reached TIMEOUT only at `3.002s`. A retryable error with `retry_after=90` and the same `.05s` lifetime likewise slept the entire 90 seconds before TIMEOUT. This contradicts the documented “hard deadline enforced during LLM and tool waits.” Check remaining lifetime before each tool/recovery sleep, cap awaits/sleep to the positive remainder, and transition directly to lifetime TIMEOUT when exhausted. Add deterministic tests for multiple tool calls and a retry delay larger than remaining lifetime.

2. **The spawn-policy snapshot is not persisted** (`src/agents/trajectory.py:101-123`). `AgentTrajectoryTurn` receives `iteration_timeout` and `max_lifetime`, but `to_dict()` omits both, so JSONL records never contain them. The current test only inspects the in-memory object. Add both fields to `to_dict()` and assert serialized/saved output.

Everything else I reviewed looks sound: per-request Codex effort is locally resolved rather than mutating the shared client; live agent-effort resolution, API pre-mutation validation/null semantics, config wiring, scheduler collection, UI/dist, and per-iteration provider/model/effort stamps are coherent.

Validation: targeted affected suite **454 passed**; targeted Ruff clean; `npm run check` clean; `git diff --check` clean. GitHub `tests`, `build`, `lint-no-new`, `types-no-new`, and `coverage-no-drop` are green.

Calmingstorm:
Re-review of `3e79722`: **one deadline blocker remains.**

The two originally reported defects are substantially addressed:

- `AgentTrajectoryTurn.to_dict()` now persists `iteration_timeout` and `max_lifetime`, and the serialized iteration stamps are covered.
- The bonus-second floor is gone, each tool await is capped to the positive remainder, each subsequent tool is guarded, and recovery sleep is capped/guarded.

However, lifetime exhaustion during the **last tool call of the last allowed iteration** is still misclassified as successful completion. `src/agents/manager.py:988-1021` catches the lifetime-capped tool `TimeoutError` as an ordinary tool timeout, records it, and transitions back to `READY`; there is no post-tool deadline check. With no next tool and no next iteration, execution falls through at lines 1030-1031 to `COMPLETED`.

Live repro against this revision:

```python
agent = AgentInfo(..., iteration_timeout=900, max_lifetime=0.05)
# LLM returns exactly one tool; tool sleeps 5s
await _run_agent(agent, "sys", [], llm, tool, max_iterations=1)
```

Observed:

```text
Agent one-tool tool slow timed out
{'state': 'completed',
 'reason': 'max iterations (1) reached',
 'elapsed': 0.0505,
 'iterations': 1}
```

The three-tool regression test misses this because the pre-call guard for tool 2 performs the transition. Add a deadline check after every tool await/result (including the final tool), record the partial iteration once, and transition directly to `TIMEOUT`. Please pin the one-tool/one-iteration shape.

Validation otherwise clean:

- targeted tests: `460 passed`
- targeted Ruff and mypy: clean
- UI templates/build: clean
- lint/type gates: `new=0`
- `git diff --check`: clean
- all GitHub checks: green

Calmingstorm:
LGTM at `7cbdb371cb0d3882b3dae7a751da75ff37636207` — the final blocker is resolved.

The new post-tool deadline check runs after the iteration is recorded and before either the `READY` transition or max-iteration completion path. I reran the exact one-tool / `max_lifetime=0.05` / `max_iterations=1` repro: it now terminates `TIMEOUT` with `lifetime exceeded`, returns in ~0.05s, and preserves the lifetime-capped tool timeout in the saved trajectory.

Final validation:
- PR-head targeted suite: **461 passed**
- Exact live repro: **TIMEOUT**, not `COMPLETED`
- Ruff: clean
- mypy: clean
- lint gate: `new=0`
- type gate: `new=0`
- `git diff --check origin/master...HEAD`: clean
- WebUI template/build check: clean
- GitHub tests, coverage, lint, types, and WebUI build: all green

No remaining blockers. **Merge-ready.**

PR #227, opened 2026-07-10
fix: supervisor-independent in-place restart for self-update and setup wizard

Calmingstorm:
## Problem

Field report from an external git-clone install: the WebUI self-update applies correctly, but the service never comes back up — it needs a manual start every time. The first-boot setup wizard has the same failure shape.

**Root cause:** `apply_update` and `/api/setup/complete` restart via `call_later(2, os.kill(getpid(), SIGTERM))` and trust the supervisor to resurrect the process. A SIGTERM shutdown exits **0** (clean), and only `Restart=always` restarts a clean exit — `Restart=on-failure`, systemd's default `Restart=no`, Docker without a restart policy, and no-supervisor setups all stay down. The git-based updater can only ever run on git-clone installs (the .deb ships no repository), which is exactly the population running hand-rolled units. The wizard's comment even claimed `Restart=on-failure` would cover it — it does not.

## Fix — in-place re-exec after genuine quiescence

Design was reviewed over the bridge before implementation; all three review blockers are addressed:

**1. Shutdown is now a completion barrier** (`src/__main__.py`) — review blocker: the old orchestration let `bot.close()` unblock `bot.start()`, so `run()` returned and `loop.close()` destroyed the still-running shutdown task mid-cleanup (reproduced: "Task was destroyed but it is pending", `health.stop()` never completing). Now: one tracked shutdown task (repeat signals reuse it — no duplicate teardown), `run()` awaits it in a `finally`, `loop.stop()` removed, and the loop is drained before close (stragglers cancelled + awaited, `shutdown_asyncgens`, `shutdown_default_executor`). This fixes a pre-existing cleanup-cutoff bug independent of restart.

**2. The real process registry is terminated** (`src/discord/wiring.py`, first commit) — review finding: `shutdown_services` read `bot.process_registry`, which nothing ever binds; `manage_process` children were never terminated by app shutdown (systemd's cgroup kill silently covered them — an in-place exec does not get that safety net). Now reads the executor's lazy `_process_registry` — the web API's seam — without instantiating an unused registry. The old tests faked the phantom attribute and validated the bug.

**3. Wizard env staleness across exec** (`src/web/api/config_admin.py`) — review blocker: exec inherits the already-loaded environment and `load_dotenv(override=False)` keeps the stale `DISCORD_TOKEN`. The fresh token rides as an exec-time env override merged by `execve`.

**Plus:**
- `src/restart.py` — process-global restart intent; `reexec()` reconstructs `python -m src` (argv passthrough; identical for the console script). **Exec failure exits nonzero** so `Restart=on-failure` gets its chance — a clean-exit fallback would recreate the exact stranding being removed.
- `self_update.py` — friendly 409 for non-git installs via `git rev-parse --is-inside-work-tree` (a linked worktree's `.git` is a *file* — ask git, not the filesystem), checked before resolving releases or mutating anything.
- `docs/configuration.md` — restart semantics; `Restart=always` still recommended for crash recovery.

## Audit notes (from the design review's 'other process owners' item)

- **MCP stdio children:** `MCPManager` has no construction site anywhere in src/ — `bot.mcp_manager` is read but never bound (the integrations API honestly 503s). Dormant; nothing to reap.
- **Ad-hoc in-flight `run_command` children during the 2s restart window:** abandoned by the old path too (cgroup kill); with exec they persist until reaped at the next full stop. Restart is a deliberate admin action at a chosen moment — documented residual rather than adding a process-group sweep. Flagging for review opinion.
- Out of scope: `update.js` reload-once-after-8s (could poll until healthy — UI follow-up), `_WS_CHAT_TIMEOUT` (unrelated, already tracked).

## Tests (+20; all exec/kill primitives stubbed — nothing real can exec, signal, or spawn)

Barrier: signal-triggered shutdown completes delayed `health.stop()` before returning · double SIGTERM = one teardown · stragglers drained before close. Exec: runs only after quiescence, argv reconstructed, env override wins over inherited · exec failure exits **1** even after clean shutdown · no flag → no exec. Endpoints: success arms the flag; every failure path leaves it unset; non-git 409 before any resolution; git-missing and `rev-parse`-false variants. Registry: real seam pinned, no lazy creation, None-executor tolerated. Autouse fixture resets restart state around every test.

## Gates

Full suite **7,338 passed / 4 skipped** · ruff **0** · mypy **0** (222 files) · lint-gate new=0 · type-gate new=0 · coverage-gate findings=0 (86.9%, `src/restart.py` enters the gated set).

No deploy, no release — PR only.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -207,3 +207,16 @@ logging:
 | Logs | `/var/log/odin/` |
 | Application | `/opt/odin/` |
 | Systemd | `/usr/lib/systemd/system/odin.service` |
+
+## Restarts (self-update and setup wizard)
+
+The WebUI self-updater (Updates page) requires a **git-clone install** —
+`.deb` installs have no repository and should upgrade via `apt` instead
+(the endpoint answers 409 with the same hint).
+
+After a successful update — and after the first-boot setup wizard saves its
+config — Odin restarts **in place** by re-executing itself once graceful
+shutdown completes. Recovery therefore does not depend on the service
+unit's `Restart=` policy, Docker restart policy, or any supervisor at all.
+`Restart=always` (what the packaged unit ships) is still recommended so the
+service also recovers from crashes and reboots.
--- a/src/__main__.py
+++ b/src/__main__.py
@@ -14,6 +14,8 @@ import signal
 import sys
 from pathlib import Path
 
+from src import restart
+
 
 def _wire_observability(health, bot, log) -> None:
     """Register Prometheus metric sources and component health checks.
@@ -124,41 +126,60 @@ def main() -> None:
     # fatal startup error (bad token, boot-time DNS failure) exited 0 and only
     # Restart=always kept the service recovering.
     exit_code = 0
+    shutdown_task: asyncio.Task[None] | None = None
+
+    def request_shutdown() -> asyncio.Task[None]:
+        """Create the shutdown task exactly once; repeat requests reuse it.
+
+        A second SIGTERM must not start a second teardown pass — every
+        cleanup step would run twice, racing its own first invocation.
+        """
+        nonlocal shutdown_task
+        if shutdown_task is None:
+            shutdown_task = loop.create_task(shutdown())
+        return shutdown_task
 
     async def run() -> None:
         nonlocal exit_code
-        await health.start()
-
-        async def _webhook_send(channel_id: str, text: str) -> None:
-            channel = bot.get_channel(int(channel_id))
-            if channel:
-                # Admin-configured webhook target; kind not enforced, so the
-                # union includes send-less channel types (Category/Forum).
-                await channel.send(scrub_response_secrets(text))  # type: ignore[union-attr]
-            else:
-                log.warning("Webhook: channel %s not found", channel_id)
-
-        health.set_send_message(_webhook_send)
-        if hasattr(bot, "scheduler") and hasattr(health, "set_trigger_callback"):
-            health.set_trigger_callback(bot.scheduler.fire_triggers)
-
-        _wire_observability(health, bot, log)
-
-        def handle_signal() -> None:
-            log.info("Shutdown signal received")
-            loop.create_task(shutdown())
-
-        for sig in (signal.SIGTERM, signal.SIGINT):
-            loop.add_signal_handler(sig, handle_signal)
-
         try:
+            await health.start()
+
+            async def _webhook_send(channel_id: str, text: str) -> None:
+                channel = bot.get_channel(int(channel_id))
+                if channel:
+                    # Admin-configured webhook target; kind not enforced, so the
+                    # union includes send-less channel types (Category/Forum).
+                    await channel.send(scrub_response_secrets(text))  # type: ignore[union-attr]
+                else:
+                    log.warning("Webhook: channel %s not found", channel_id)
+
+            health.set_send_message(_webhook_send)
+            if hasattr(bot, "scheduler") and hasattr(health, "set_trigger_callback"):
+                health.set_trigger_callback(bot.scheduler.fire_triggers)
+
+            _wire_observability(health, bot, log)
+
+            def handle_signal() -> None:
+                log.info("Shutdown signal received")
+                request_shutdown()
+
+            for sig in (signal.SIGTERM, signal.SIGINT):
+                loop.add_signal_handler(sig, handle_signal)
+
             health.set_ready(True)
             log.info("Connecting to Discord…")
             await bot.start(config.discord.token)
         except Exception as exc:
             exit_code = 1
             log.error("Fatal error: %s", exc, exc_info=True)
-            await shutdown()
+        finally:
+            # Completion barrier: run() returns only after teardown actually
+            # finished. shutdown()'s bot.close() unblocks bot.start() above,
+            # so without this await run_until_complete() would return — and
+            # main() would close the loop — while the shutdown task was still
+            # mid-cleanup ("Task was destroyed but it is pending"), silently
+            # skipping whatever remained (health stop, session saves).
+            await request_shutdown()
 
     async def shutdown() -> None:
         log.info("Shutting down…")
@@ -192,33 +213,62 @@ def main() -> None:
             await health.stop()
         except Exception:
             log.exception("health stop error")
-        loop.stop()
 
     try:
         loop.run_until_complete(run())
     except (KeyboardInterrupt, SystemExit) as exc:
         # Ctrl-C stays a clean stop; an intentional SystemExit keeps its
-        # original code rather than being normalized.
+        # original code rather than being normalized. run()'s finally has
+        # normally completed shutdown already; a raw KeyboardInterrupt that
+        # broke out of the loop itself still needs it finished here.
         if isinstance(exc, SystemExit) and exc.code:
             exit_code = exc.code if isinstance(exc.code, int) else 1
-        loop.run_until_complete(shutdown())
+        loop.run_until_complete(request_shutdown())
     except asyncio.CancelledError:
         # Cancellation reaching the top level is a shutdown path, not a
         # fatal error — exit clean unless a fatal path already set a code.
-        loop.run_until_complete(shutdown())
+        loop.run_until_complete(request_shutdown())
     except Exception:
-        # Failures outside run()'s own guard (e.g. the health server's port
-        # bind) previously tracebacked out with no clean shutdown. Cleanup
-        # failures are logged separately and never mask the fatal code.
+        # Failures outside run()'s own guard previously tracebacked out with
+        # no clean shutdown. Cleanup failures are logged separately and never
+        # mask the fatal code.
         exit_code = 1
         log.exception("Fatal error during startup")
         try:
-            loop.run_until_complete(shutdown())
+            loop.run_until_complete(request_shutdown())
         except Exception:
             log.exception("Cleanup after fatal startup error failed")
     finally:
+        # Drain before close: cancel stragglers, then run the loop's async
+        # generator and default-executor shutdown hooks. Closing without this
+        # destroys still-pending work mid-await — and an in-place restart
+        # would exec over half-finished writes.
+        try:
+            pending = [t for t in asyncio.all_tasks(loop) if not t.done()]
+            for task in pending:
+                task.cancel()
+            if pending:
+                loop.run_until_complete(
+                    asyncio.gather(*pending, return_exceptions=True)
+                )
+            loop.run_until_complete(loop.shutdown_asyncgens())
+            loop.run_until_complete(loop.shutdown_default_executor())
+        except Exception:
+            log.exception("Event-loop drain failed")
         loop.close()
         log.info("Odin stopped")
+
+    if restart.restart_requested():
+        # In-place restart (self-update / setup wizard): replace the process
+        # image instead of exiting so recovery does not depend on the unit's
+        # Restart= policy. Exec failure exits nonzero — a clean-exit fallback
+        # would recreate exactly the stranding this path removes.
+        log.info("Restart requested — re-executing in place")
+        try:
+            restart.reexec()
+        except OSError:
+            log.exception("In-place restart failed")
+            sys.exit(exit_code or 1)
     if exit_code:
         sys.exit(exit_code)
 
--- a/src/discord/wiring.py
+++ b/src/discord/wiring.py
@@ -10,8 +10,9 @@ slash commands, the first system-prompt build) stays in ``OdinBot.__init__``.
 
 ``shutdown_services(bot)`` is the teardown mirror, moved verbatim from
 ``OdinBot.close()``. It takes the bot rather than a ``BotServices`` because
-several torn-down attributes are late-bound on the bot (``health_server``,
-``process_registry``) or reached via property alias (``knowledge``), and
+several torn-down attributes are late-bound on the bot (``health_server``),
+lazily created on the executor (``_process_registry``), or reached via
+property alias (``knowledge``), and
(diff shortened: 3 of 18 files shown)


Calmingstorm:
Re-reviewed at `863c551`. **One blocking issue remains: the audit-note subprocess residual is not safe to accept for an in-place exec.**

`src/__main__.py:247-253` cancels all remaining asyncio tasks immediately before exec. The ad-hoc command owners do not turn cancellation into child cleanup:

- `src/tools/ssh.py:104-121` and `177-238` clean up only `TimeoutError`; `asyncio.CancelledError` bypasses those handlers, leaving the local shell/SSH process alive.
- `src/tools/process_manager.py:64-70,141-149` terminates only the shell PID. It does not create or terminate a process group, so descendants can survive even though the newly corrected registry seam invokes `shutdown()`.

I reproduced both against this HEAD:

```text
run_local_command task cancelled:
{'pid': 3024653, 'alive_after_task_cancellation': True}

ProcessRegistry.shutdown() after a managed shell spawned `sleep 30`:
{'grandchild_pid': 3027168, 'alive_after_registry_shutdown': True}
```

This is worse than the audit note says. The process is not necessarily “reaped at the next full stop”: after exec, the new image has neither the old task nor its `ProcessRegistry` record, so a no-supervisor install may never reap it; only an eventual external cgroup/process-tree stop happens to do so. The command can also continue mutating state concurrently with the fresh Odin process.

Please make the subprocess ownership boundary cancellation-safe rather than adding a broad PID sweep:

1. Start local shell-owned commands in their own session/process group.
2. On timeout **and cancellation**, terminate the group, wait boundedly, escalate to kill, and reap it; re-raise `CancelledError`.
3. Apply the same direct-child kill/reap discipline to spawned SSH clients (remote-command persistence beyond a severed SSH connection can remain documented as the pre-existing remote caveat).
4. Make `ProcessRegistry.shutdown()` terminate the managed process group, not only its shell leader.
5. Pin both shapes: cancelling an in-flight local `run_command` leaves no descendant, and registry shutdown leaves no descendant. The tests should assert the child PID is gone, not merely that the asyncio reader task is done.

Everything else in the restart implementation reviewed coherently: the single shutdown task and completion barrier, loop drain, nonzero exec failure, reconstructed argv/env override, restart intent isolation, worktree-aware git detection before mutation, and the corrected lazy registry seam.

Validation on this HEAD:

- Focused tests: **86 passed**
- Targeted Ruff: clean
- Targeted mypy: clean
- Lint/type gates: `new=0`
- `git diff --check`: clean
- All GitHub checks: green

Calmingstorm:
Re-review at `a59ad06`: **one process-tree blocker remains.**

The original cancellation/timeout and registry leaks are fixed for ordinary descendants, the own-process-group guard is sound, SSH retry clients are now reaped before the next attempt, and both of the original repros pass. However, `terminate_process_tree()` returns as soon as the **group leader** exits after SIGTERM (`src/tools/ssh.py:89-94`). It does not verify that the process group is empty before returning. A descendant can ignore SIGTERM while the shell leader exits normally from that same TERM, leaving the descendant alive across re-exec.

Live repro against this HEAD:

```text
command tree:
  session-leading shell
    child shell: trap "" TERM; sleep 30

{'leader': 3043224, 'child': 3043225,
 'leader_pgid': 3043224, 'child_pgid': 3043224}

terminate_process_tree(proc, grace=.3) ->
{'leader_returncode': -15,
 'stubborn_descendant_alive_after_helper': True}
```

The existing escalation test does not catch this shape because `trap "" TERM; sleep 30` makes the **leader itself** ignore TERM, forcing the helper down the SIGKILL arm. The missed case is a TERM-compliant leader with a TERM-immune descendant.

Required adjustment:

1. Capture whether the child owns its process group before signalling.
2. After the TERM grace, do not use leader exit alone as proof that a managed group is gone. Probe the captured group independently and SIGKILL it if any member remains, even when `proc.wait()` already completed.
3. Continue to wait/reap the direct child boundedly; retain child-only behavior for SSH/same-group processes.
4. Pin the exact compliant-leader / TERM-immune-descendant repro above and assert the descendant is gone.

Everything else in the blocker fix reviewed cleanly.

Validation:
- Focused restart/process/SSH suite: **287 passed**
- Ruff: clean
- mypy: clean
- `git diff --check`: clean
- All GitHub checks: green

Calmingstorm:
Re-reviewed `93ff0ac`. The compliant-leader / TERM-immune-descendant case **is fixed when cleanup begins while the leader is still alive**, but one process-tree blocker remains.

`terminate_process_tree()` still returns immediately at `src/tools/ssh.py:77-78` when `proc.returncode is not None`. That is unsafe for the same leaderless-group case: an independently exiting shell leader can already be reaped while a descendant retains the command's stdout pipe, leaving `run_local_command()` blocked in `communicate()`. Cancellation during restart then calls the helper, which exits before discovering or signalling the surviving group.

I reproduced the actual `run_local_command()` cancellation shape on this HEAD with a naturally exiting session leader and a TERM-immune descendant:

```text
{'child': 3049706,
 'exists_after_cancel': True,
 'status': ['State:\tS (sleeping)', 'PPid:\t1', 'NSpgid:\t3049705']}
```

A direct helper repro is even smaller: spawn a new-session shell which backgrounds `sleep 20` and exits 0, wait for the leader first, then call `terminate_process_tree()`:

```text
{'leader': 3050004,
 'leader_returncode_before_cleanup': 0,
 'descendant': 3050005,
 'descendant_pgid': 3050004,
 'descendant_alive_after_cleanup': True}
```

The new test does not cover this boundary because it invokes the helper before sending TERM, while `proc.returncode` is still `None`.

**Required adjustment:** preserve the owned process-group identity independently of leader liveness and allow cleanup to signal that known group after the leader has already exited. Do not infer group ownership generically after exit; the SSH path intentionally shares Odin's group. A safe shape is an explicit known PGID / `owns_group` argument captured for `start_new_session=True` local and ProcessRegistry spawns, while SSH remains child-only. Pin both:

1. helper called after a new-session leader has already exited but its descendant remains;
2. `run_local_command()` cancellation while `proc.returncode` is already set and a descendant still holds stdout open.

The `93ff0ac` behavior itself is otherwise correct: PGID is snapshotted before TERM, the group is probed independently, and a surviving group is SIGKILLed even after the leader exits in response to TERM.

Validation on `93ff0ac`:

- focused restart/process/API suite: **123 passed**
- new compliant-leader test: **5/5 repeated passes**
- Ruff: clean
- mypy: clean
- `git diff --check`: clean
- live original compliant-leader repro: descendant killed (brief zombie, then reaped)
- GitHub lint/types green; tests/coverage were still running at review time

Calmingstorm:
Re-review at `3bb1a8a`: **the broadcast guard is sound for the dangerous `killpg` targets, and the `owned_pgid` path fixes the leader-already-reaped case in `run_local_command`, but PR #227 is not merge-ready yet.**

Two blockers remain:

1. **`ProcessRegistry` still leaks a leaderless owned group when the shell exits naturally.**

   `ProcessRegistry.shutdown()` only calls `kill()` for entries whose status is `running`, and `kill()` itself refuses completed entries. `_read_output()` marks the entry completed as soon as a shell exits and its output closes. A shell that backgrounds a redirected child therefore leaves the child alive, records itself completed, and is skipped during restart shutdown.

   Repro against this HEAD:

   ```text
   before_shutdown:
     leader_returncode: 0
     registry_status: completed
     child_pid: 3093506
   shutdown_killed_count: 0
   descendant_alive_after_shutdown: true
   ```

   Command shape:

   ```sh
   sleep 20 >/dev/null 2>&1 & echo $! > PIDFILE
   ```

   Do not solve this by group-signalling every old `completed` entry during shutdown; those entries can remain for an hour, making stale-PGID/PID reuse a real unrelated-process kill risk. Clean up the owned group when the managed leader is observed exiting, while ownership is fresh, or otherwise retain a safe live ownership mechanism. Pin the naturally exited leader + redirected live descendant through the real registry path.

2. **The new direct helper test is racy and has already failed the required coverage gate.**

   `TestLeaderAlreadyExited.test_helper_kills_group_after_leader_reaped` does `await proc.wait()` while the background `sleep` inherits the subprocess stdout pipe. On this host that wait takes the full 30 seconds; the descendant is no longer reliably alive when the test reaches `os.kill(stubborn, 0)`. In GitHub's coverage job it was already gone, producing:

   ```text
   FAILED ...test_helper_kills_group_after_leader_reaped
   ProcessLookupError: [Errno 3] No such process
   1 failed, 7359 passed, 5 skipped
   ```

   Make the setup deterministic: either poll until `proc.returncode` is set without awaiting pipe EOF, or give the leader/descendant `DEVNULL` output so `proc.wait()` can complete while the descendant demonstrably remains alive. Then rerun the coverage gate.

On the two requested concerns specifically:

- `_is_signallable_group()` plus `pgid == proc.pid` correctly rejects `None`, non-integers, `<=1`, and Odin's own group before every `killpg`; unsafe values fall back to child-only signalling. `ProcessKillTool` rejecting `pid <= 0` is also correct.
- `owned_pgid=proc.pid` closes the round-3 leaderless-group hole for the local command paths. Both the direct helper behavior and the real `run_local_command()` cancellation path passed locally; the latter is the operationally meaningful proof.
- The guard does not cryptographically prove group ownership after leader death or eliminate arbitrary foreign-PGID reuse; only a stronger lifetime primitive such as a dedicated cgroup could do that. Given immediate cleanup and a by-construction PGID, that narrow race is reasonable. It is **not** reasonable for hour-old completed registry records, hence blocker 1.

Validation:

- Focused suite: **135 passed**
- Round-3 and broadcast cases: **5 repeated rounds passed**
- Ruff: clean
- mypy: clean
- lint/type gates: `new=0`
- `git diff --check`: clean
- GitHub tests/lint/types: green
- GitHub coverage: **failed on the new leader-exited test**

Calmingstorm:
Re-review at `76baadd`: **the behavior scope is intended, and the prior two findings are fixed in isolation, but one shutdown race remains blocking.**

### Behavior ruling

Yes: `manage_process` should own the full process tree for the command it starts. When the managed leader completes, its surviving group descendants should be terminated during normal operation, not only during shutdown. Allowing `cmd &` to escape would create an untracked process that cannot subsequently be polled, killed, lifetime-limited, or reliably reaped before in-place exec. A genuinely persistent helper should be started as its own managed process or external service.

### Confirmed fixes

- `_read_output()` now calls `terminate_process_tree(..., owned_pgid=proc.pid)` while the completed entry's group ownership is fresh. The original redirected-descendant repro is closed: the descendant was gone, the record terminal, and `shutdown_killed == 0`. The new test passed **10/10** repeated runs.
- `TestLeaderAlreadyExited.test_helper_kills_group_after_leader_reaped` now uses `DEVNULL`, so `proc.wait()` reaps the leader without waiting for a descendant-held pipe. That setup is deterministic and correctly pins the helper boundary.

### Remaining blocker: shutdown can cancel the new reap in progress

`ProcessRegistry._read_output()` marks the record terminal at `src/tools/process_manager.py:237-240`, then awaits the new TERM/grace/KILL group reap at `:251-254`. During that await, `shutdown()` skips the record because it is no longer `running`, then cancels the still-active reader task at `:190-199`. `terminate_process_tree()` intentionally propagates cancellation, so a descendant which survives TERM can escape before the SIGKILL phase. This is exactly the restart window the PR needs to close.

Live repro against `76baadd` used a naturally exiting managed shell with a redirected Python descendant that ignores SIGTERM. Shutdown ran after the record became `completed` but while its reader/reaper task was still active:

```text
pre-shutdown {
  'status': 'completed',
  'reader_done': False,
  'child_alive': True
}
post-shutdown {
  'shutdown_killed': 0,
  'reader_done': True,
  'reader_cancelled': True,
  'child_alive': True
}
```

Please make terminal publication and group cleanup one lifecycle operation from shutdown's perspective. Acceptable shapes include keeping the record `running` until group reap finishes, or representing/awaiting a dedicated cleanup state/task which shutdown does not cancel prematurely. The invariant to pin is:

1. leader has exited;
2. redirected descendant ignores TERM and remains alive during the grace period;
3. registry record/group cleanup has started;
4. `shutdown()` runs immediately;
5. shutdown does not return until the descendant is gone and cleanup is no longer pending.

Avoid merely shielding an orphan cleanup task unless shutdown explicitly awaits it; in-place exec must not replace the process while that task is unfinished.

Validation on this HEAD:

- Focused process/restart suite: **75 passed**
- New leaderless-group test: **10/10 repeated passes**
- Ruff: clean
- mypy: clean
- `git diff --check`: clean
- All GitHub checks: green

Calmingstorm:
Re-reviewed at `c098840` — **LGTM. The round-5 shutdown race is closed.**

The lifecycle now has the ordering the in-place exec path requires:

- running leaders/groups are terminated first;
- terminal records are no longer treated as proof that their reader/reaper has finished;
- every extant reader/reaper task is awaited under the bounded reap timeout;
- only the explicit wedged-task fallback cancels a reaper, and the top-level loop drain still gathers cancellation before `reexec()`.

I reran the original real-process race shape 10 times: terminal record, reader mid-reap, TERM-immune descendant still alive at shutdown entry. All 10 runs returned with the reader completed rather than cancelled and the descendant gone. The focused modeled regression also passed 10/10.

The previously approved ownership semantics remain intact: a managed process group is reaped whenever its leader exits, not only during shutdown.

Validation on this exact HEAD:

- focused restart/process/shutdown tests: **117 passed**
- original race repro: **10/10 passed**
- modeled race regression: **10/10 passed**
- Ruff: clean
- mypy: clean
- `git diff --check`: clean
- GitHub tests, coverage, lint, and types: all green

No remaining review blocker. Clean for the requested branch deployment and live restart test. The process tree has finally stopped trying to achieve immortality through shell semantics.

PR #228, opened 2026-07-10
feat(image): native OpenAI image generation, toggleable with ComfyUI

Calmingstorm:
## Summary

Adds native OpenAI image generation as a toggleable alternative to the ComfyUI image tool, behind a backend-neutral `ImageBackend` seam. The old `generate_image` ComfyUI tool becomes one backend rather than a permanent special case.

**Feasibility proven live** before building: the `image_generation` Responses tool works over our existing Codex ChatGPT OAuth backend (`chatgpt.com/backend-api/codex/responses`) using the current pool account — HTTP 200, full SSE flow, a real PNG returned. Subscription-quota-backed (no per-token billing); image gen draws on the same account usage limit as chat.

## Design

- **`src/tools/image/`** — `ImageBackend` ABC + `ImageResult` + typed error taxonomy (`pre_generation` flag = safe-to-fail-over/fall-back); `OpenAIImageBackend` (native wire impl), `ComfyUIImageBackend` (wraps the existing client), `ImageBackendSelector`.
- **Native backend** rides the **same `CodexAuthPool`** the chat client uses (no separate auth) but has its own transport, SSE parser, and circuit breaker so image failures never poison chat. Failover only on pre-generation failures; **once a POST returns 200 the account is pinned — no retry/rotation/fallback** (never double-spends quota). partial_image discarded; one terminal PNG required; base64 length / decoded size / PNG magic all bounded; no token/account/payload in logs, errors, or results.
- **Modes** (`image.backend`): `auto` (default) follows the active provider — codex → native (ComfyUI as pre-generation fallback), otherwise ComfyUI; `openai` / `comfyui` force one. `outer_model` pinned in config (not the chat model), sizes allowlisted to probed values, kill switch.
- **Visibility** is structural: the tool appears iff a native (Codex provider + creds) or ComfyUI backend is configured — **Kimi with no ComfyUI hides it**. A live provider switch rebuilds the registry.

## Two deliberate deviations from the design review — Aaron's explicit calls

1. **Availability IS tied to the active chat provider** (Odin's R2 argued to decouple). Rationale: native image rides *whatever Codex auth Odin is currently using*, so on Kimi there's no active Codex auth → ComfyUI only.
2. **`auto` is the flat default with no migration guard** (Odin flagged a silent behavior change for existing ComfyUI installs). Intended: existing installs switch to native-first on upgrade.

## Gates

7,406 passed / 4 skip / 0 failed · ruff 0 · mypy 0 (227 files) · lint/type/coverage gates all clean. New fixture tests cover failure classification, the no-retry-after-generation boundary, the >512KB SSE line, the visibility matrix, and selector routing.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -251,6 +251,22 @@ comfyui:
   enabled: true
   url: http://localhost:8188
 
+# Image generation backend selection.
+#   auto    (default): follow the active chat provider — on 'codex' use native
+#            OpenAI (ComfyUI is the pre-generation fallback); on any other
+#            provider use ComfyUI only. If neither is available the tool hides.
+#   openai : native OpenAI only (rides the current Codex auth; Codex-provider only).
+#   comfyui: ComfyUI only.
+image:
+  backend: auto
+  openai:
+    enabled: true          # kill switch for the native wire implementation
+    outer_model: gpt-5.5   # Responses model hosting the image tool (pinned, not the chat model)
+    image_model: gpt-image-2
+    # Only sizes probed against this private endpoint — widen after verifying each.
+    allowed_sizes: ["1024x1024"]
+    default_size: "1024x1024"
+
 # Web UI and API — all web configuration lives here (not in .env)
 web:
   enabled: true
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -141,6 +141,47 @@ Leave `cdp_url` empty to launch a local headless Chromium. Set to `ws://host:por
 
 Run `playwright install chromium` after installation.
 
+## Image Generation
+
+```yaml
+comfyui:
+  enabled: true
+  url: http://localhost:8188
+
+image:
+  backend: auto            # auto | openai | comfyui
+  openai:
+    enabled: true          # kill switch for the native wire implementation
+    outer_model: gpt-5.5   # Responses model hosting the image tool (pinned)
+    image_model: gpt-image-2
+    allowed_sizes: ["1024x1024"]
+    default_size: "1024x1024"
+```
+
+The `generate_image` tool can target two backends:
+
+- **Native OpenAI** — generates via the `image_generation` tool on the Codex
+  ChatGPT OAuth backend, riding the **same account Odin uses for chat** (no
+  separate auth; subscription-quota-backed, so it draws on that account's usage
+  limit). Available only while the active provider is `codex`.
+- **ComfyUI** — the local Stable-Diffusion backend (`comfyui.enabled`), which
+  also supports `negative`, `width`/`height`, and checkpoint `model`.
+
+`backend` selects between them:
+
+- `auto` (default) follows the active chat provider: on `codex`, native OpenAI
+  is used with ComfyUI as a pre-generation fallback; on any other provider,
+  ComfyUI only. A `negative`/`width`/`height`/`model` argument routes the
+  request to ComfyUI. If neither backend is available (e.g. Kimi with no
+  ComfyUI) the tool is hidden from the registry.
+- `openai` forces native OpenAI (ComfyUI-only arguments are rejected).
+- `comfyui` forces ComfyUI.
+
+`outer_model` is pinned here rather than following your chat model, so changing
+the chat model (Sol/Terra/…) never alters image generation. Only sizes listed in
+`allowed_sizes` are accepted — widen it only after verifying a size works against
+this private endpoint.
+
 ## Web Management UI
 
 ```yaml
--- a/src/config/schema.py
+++ b/src/config/schema.py
@@ -596,6 +596,46 @@ class ComfyUIConfig(BaseModel):
     default_checkpoint: str = ""
 
 
+class ImageOpenAIConfig(BaseModel):
+    """Native OpenAI image generation over the Codex ChatGPT OAuth backend.
+
+    Rides the SAME CodexAuthPool / current account Odin uses for chat — no
+    separate auth, no per-token API billing (subscription-quota-backed). The
+    outer model is pinned here rather than inherited from the chat model so a
+    Sol/Terra/UI change can't silently alter image generation. Both models are
+    config-only allowlisted, never arbitrary strings from the tool call.
+    """
+
+    enabled: bool = True  # kill switch for the native wire implementation
+    outer_model: str = "gpt-5.5"  # Responses model that hosts the image tool
+    image_model: str = "gpt-image-2"  # the image_generation tool's model
+    # Only sizes actually probed against THIS private endpoint are allowed;
+    # widen after verifying each value (the Codex endpoint != public Images API).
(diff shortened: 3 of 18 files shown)


Calmingstorm:
**PR #228 review at `015f106` — changes requested.** The two Aaron-directed deviations are accepted as intentional product behavior, and the post-200 pinning itself is sound, but the native path has several implementation blockers.

1. **The real `CodexAuthPool` cannot generate an image.**

   `src/llm/codex_auth.py` exposes `account_count` as a property, but `src/tools/image/openai_backend.py:114` calls it as `self._auth.account_count()`. The fixture pool defines a method, so the tests conceal the production mismatch. A real-pool repro fails before its first POST:

   ```text
   TypeError: 'int' object is not callable
   ```

   Use the real pool contract and pin a test with `CodexAuthPool` or a property-faithful fake.

2. **The image backend does not remain attached to the live shared auth pool.**

   `src/discord/wiring.py:630-639` snapshots `services.codex_client.auth` once. If Odin starts without a Codex client and credentials are added through the existing live reload, the selector permanently has no native backend. If Codex is disabled and later recreated, images retain the stale old pool while chat uses the replacement. This violates the stated same-pool contract and makes the provider/config-driven visibility stale.

   Make auth/backend resolution late-bound through `llm_gateway.codex_client`, or atomically rebind it whenever Codex reload creates/replaces/removes the client. Tool visibility must also reflect an actually configured live pool, not merely `openai_codex.enabled`; currently enabled Codex plus a missing credentials file advertises a tool that can only return “not available.” Pin boot-without-auth → live login/reload, pool replacement, and missing-credentials visibility.

3. **Pre-generation failure classification is incomplete, so `auto` fallback is not reliable.**

   - `CodexAuthPool.acquire()` can raise `RuntimeError` when every account is cooling down or token acquisition fails. That escapes the `ImageGenError` taxonomy, bypasses ComfyUI fallback, and can surface pool/token-refresh details through the tool handler.
   - `CircuitBreaker.check()` raises `CircuitOpenError`, also outside the taxonomy, so an open image breaker bypasses `auto` fallback.
   - Other 4xx request/policy failures call `breaker.record_failure()` at `openai_backend.py:146`. Request-specific errors were explicitly meant not to poison global backend health; three bad requests can open the image breaker for everyone.

   Wrap pool exhaustion/acquisition failures and an open breaker as sanitized pre-generation image errors, while preserving cancellation. Do not count request/content-policy failures against the transport breaker. Pin each selector-level fallback/non-fallback outcome.

4. **The SSE size bound does not bound the SSE buffer.**

   `_read_stream()` appends to `buf` until a newline appears, but checks `max_b64` only after parsing a complete JSON event and extracting `final_b64`. An unterminated or malicious response can therefore consume arbitrary memory. I fed **32 MiB** through it with `max_image_bytes=1 KiB`; it consumed the whole stream before returning `no image returned`.

   Enforce a bounded maximum frame/buffer length while accumulating, before newline/JSON parsing, and pin an oversized unterminated frame.

5. **The native backend’s HTTP session is never closed.**

   `OpenAIImageBackend.close()` exists, but the backend is retained only through `MediaTools.image_selector`; `shutdown_services()` never reaches it. After the first native request, its `aiohttp.ClientSession` survives shutdown/re-exec cleanup. Give the selector/media owner a close path and await it during shutdown, with a lifecycle test.

**Requested focus:** the strict **no-failover after HTTP 200** boundary is correctly enforced for stream failures, terminal `response.failed`, malformed output, and size/PNG rejection. `partial_image` is discarded as intended. The failures above are all before/around that boundary, not an objection to pinning after acceptance.

**Product deviations:** approved as explicit Aaron decisions:

- Native availability may follow the active provider; Kimi without ComfyUI hides the tool.
- `auto` may be the flat default and intentionally switch existing ComfyUI installs to native-first.

Validation performed:

- Focused image/media/gateway/parity tests: **100 passed**
- Focused Ruff: clean
- Focused mypy: clean
- `git diff --check`: clean
- All GitHub checks: green

Calmingstorm:
**PR #228 re-review at `7288a23` — one blocker remains.**

The five reported fixes are present and work in the covered paths:

- the real `CodexAuthPool.account_count` property is used correctly;
- auth is resolved from the gateway at call time;
- pool exhaustion/open breaker are pre-generation fallback errors, while request 4xx no longer poison the breaker;
- the unterminated SSE buffer is bounded;
- the native session is reached and closed during shutdown.

However, the documented hard **no-failover-after-HTTP-200** boundary is still not airtight.

`src/tools/image/openai_backend.py:134-170` wraps the entire `async with session.post(...)` in the broad pre-generation `except (TimeoutError, aiohttp.ClientError)`. That catch also covers `resp.__aexit__()`. If a 200 response successfully yields the final image but response cleanup then raises an `aiohttp.ClientError`, execution classifies it as pre-response transport failure and retries another account.

Reproduced against this exact HEAD with a successful 200/SSE response whose `__aexit__` raises `ClientConnectionError`, followed by a second successful account:

```text
{'returned': 'openai', 'posts': 2, 'breaker_failures': 0}
```

That can generate twice after the account was supposed to be pinned.

**Required adjustment:** track acceptance independently of the outer context-manager cleanup, or narrow the pre-response catch so any timeout/client error after status 200 becomes `ImageTransportError(..., pre_generation=False)` and is never continued to another account or ComfyUI. Pin the successful-stream + failing-`__aexit__` case and assert exactly one POST.

Everything else in this re-review is clean.

Validation:

- focused image/shutdown/media/gateway/parity tests: **130 passed**
- Ruff: clean
- mypy: clean
- lint/type gates: `new=0`
- `git diff --check`: clean
- all GitHub checks: green

Calmingstorm:
**PR #228 re-review at `ad44f00`: LGTM — no remaining blocker.**

The `committed` boundary closes the response-cleanup hole correctly:

- It flips immediately when HTTP 200 is observed, before stream consumption.
- A later `TimeoutError`/`aiohttp.ClientError`, including one raised by `__aexit__()` after a valid image was parsed, becomes `ImageTransportError(pre_generation=False)`.
- That exception leaves the account pinned: there is no second POST, account rotation, or ComfyUI fallback.
- The cleanup failure does not increment the image breaker; pre-response transport failures still do.
- Mid-stream failures remain handled by `_read_stream()` as post-generation failures.

I reran the exact successful-stream/failing-`__aexit__` shape independently and observed:

```text
{'type': 'ImageTransportError', 'pre_generation': False, 'posts': 1, 'breaker_failures': 0}
```

Validation at this HEAD:

- Focused image/media/shutdown/gateway/parity tests: **131 passed**
- Ruff: clean
- mypy for the changed backend: clean
- `git diff --check`: clean
- GitHub lint and types checks: green; tests and coverage were still running when reviewed

The previously accepted product decisions remain intentional: native availability follows the active chat provider, and flat `auto` is native-first without a ComfyUI migration guard.

**Clean to merge once the remaining GitHub checks finish green.**

PR #229, opened 2026-07-11
fix(image): native is square-only; route by capability; hide backend name

Calmingstorm:
## Problem

In Aaron's live test, "generate an image of a cat" went to ComfyUI, not native OpenAI. Root cause: the LLM auto-filled `width`/`height` from the tool schema, and the selector treated **any** width/height as a ComfyUI-only feature — short-circuiting to ComfyUI before native was tried. That over-implemented the design (only `negative`/checkpoint were meant to force ComfyUI).

Probing the endpoint also revealed it **ignores the requested size**: `1024x1024`, `1536x1024`, and `1024x1536` all returned the same ~`1254x1254` square. Native is a square-only, backend-selected-resolution generator.

## Fix (Odin-reviewed design)

- Only `negative` or a checkpoint `model` force ComfyUI. `width`/`height` fold into a canonical `size` (explicit `size` wins; one-sided/malformed rejected before routing).
- **Capability routing**: a non-square size is a ComfyUI request. In `auto` it goes straight to ComfyUI and **never falls back to native** (which would return the wrong shape); with no ComfyUI it errors honestly. In `openai`, a non-square size or a ComfyUI-only field is **rejected before acquiring auth**.
- The native payload no longer sends the ignored `size`; the backend also refuses non-square as defense in depth.
- Drop `allowed_sizes`/`default_size` (meaningless — the endpoint ignores size).
- Remove `width`/`height` from the LLM-visible tool schema (still accepted internally for replay/scheduled calls); document `size` semantics.
- The result message **drops the backend name** and reports **decoded** dimensions (`"Image generated (1254x1254, N KB) and posted"`); backend/fallback recorded in logs only.

## Routing table

| Mode | Request | Route |
|---|---|---|
| `auto` | `negative` or checkpoint `model` | ComfyUI |
| `auto` | explicit non-square size | ComfyUI directly (no native fallback) |
| `auto` | no size or square size | native first; pre-generation fallback to ComfyUI |
| `openai` | non-square size or ComfyUI-only field | reject before auth |
| `openai` | no size or square size | native |
| `comfyui` | any valid request | ComfyUI |

## Gates

7,425 passed / 4 skip / 0 failed · ruff 0 · mypy 0 (227 files) · lint/type/coverage clean. Tests pin: width/height fold to native, auto non-square → ComfyUI (zero native calls), no native fallback when ComfyUI can't do non-square, explicit-openai non-square rejection, square native routing, one-sided rejection, decoded dims.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -253,19 +253,20 @@ comfyui:
 
 # Image generation backend selection.
 #   auto    (default): follow the active chat provider — on 'codex' use native
-#            OpenAI (ComfyUI is the pre-generation fallback); on any other
-#            provider use ComfyUI only. If neither is available the tool hides.
-#   openai : native OpenAI only (rides the current Codex auth; Codex-provider only).
+#            OpenAI for square/unspecified requests (ComfyUI is the pre-generation
+#            fallback); a non-square size, a negative prompt, or a checkpoint
+#            routes to ComfyUI; on any other provider use ComfyUI only. If
+#            neither is available the tool hides.
+#   openai : native OpenAI only (rides the current Codex auth; Codex-provider
+#            only; square output only — non-square requests are rejected).
 #   comfyui: ComfyUI only.
+# Native OpenAI ignores the requested size and returns a backend-selected square.
 image:
   backend: auto
   openai:
     enabled: true          # kill switch for the native wire implementation
     outer_model: gpt-5.5   # Responses model hosting the image tool (pinned, not the chat model)
     image_model: gpt-image-2
-    # Only sizes probed against this private endpoint — widen after verifying each.
-    allowed_sizes: ["1024x1024"]
-    default_size: "1024x1024"
 
 # Web UI and API — all web configuration lives here (not in .env)
 web:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -154,8 +154,6 @@ image:
     enabled: true          # kill switch for the native wire implementation
     outer_model: gpt-5.5   # Responses model hosting the image tool (pinned)
     image_model: gpt-image-2
-    allowed_sizes: ["1024x1024"]
-    default_size: "1024x1024"
 ```
 
 The `generate_image` tool can target two backends:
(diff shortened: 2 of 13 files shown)


Calmingstorm:
PR #229 review at `ceee65e`: **changes requested**.

The square-only routing table itself is implemented correctly: legacy square `width`/`height` reaches native, non-square requests bypass native, explicit OpenAI rejects capability mismatches before calling the native backend, the native wire payload omits `size`, and the user-facing message hides the backend and uses `ImageResult` dimensions.

Two blockers remain:

1. **The new canonical validation and ComfyUI's actual capability disagree, so accepted explicit sizes are silently changed.**

`parse_size()` accepts every positive dimension through 4096 (`src/tools/image/base.py:26-46`), but `ComfyUIImageBackend._parse_size()` silently clamps each axis to 64..2048 (`src/tools/image/comfyui_backend.py:20-39`). Since non-square requests now route directly to ComfyUI, valid-at-the-selector requests do not necessarily preserve the requested size or even aspect ratio:

```text
1x2       -> 64x64
32x64     -> 64x64
64x4096   -> 64x2048
4096x1024 -> 2048x1024
```

That contradicts the tool/docs claim that ComfyUI honors exact `size` dimensions and the design requirement to reject unsupported bounds before routing. Use one canonical parser/capability contract end-to-end and reject dimensions ComfyUI cannot honor; do not clamp an already validated explicit request. Also wrap legacy `width`/`height` conversion: `width="bad"` currently escapes `_resolve_size()` as raw `ValueError`, while fractional values are silently truncated (`1.5x2.5 -> 1x2`). Malformed legacy input should become `ImageRequestError` before routing.

2. **The result is not guaranteed to report decoded dimensions for ComfyUI.**

At `src/tools/image/comfyui_backend.py:73-74`, failed PNG decoding falls back to the requested `(w, h)`. A non-PNG response is therefore returned as `mime="image/png"`, posted as `generated.png`, and reported with guessed request dimensions. The final design explicitly requires decoded—not requested—dimensions. Reject ComfyUI output when `png_dimensions()` fails rather than substituting request dimensions, and pin that case.

Validation performed:

- Focused tests: **90 passed**
- Focused Ruff: clean
- Focused mypy: clean
- `git diff --check`: clean
- All four GitHub checks: green

No other blocker found in the routing/fallback or native-payload changes.

Calmingstorm:
Re-reviewed at `8d0708f` — **LGTM. No remaining blocker.**

Both round-two issues are closed:

- `parse_size()` is now the single `64..2048` dimension contract used by selector routing and ComfyUI. Accepted requests are no longer silently clamped, including aspect-ratio-changing cases. Boundary values pass; out-of-range, fractional, malformed, and legacy non-integer inputs are rejected cleanly before backend dispatch.
- ComfyUI now requires image bytes that pass the shared PNG validation/dimension extraction and reports dimensions from that output. It no longer substitutes requested dimensions for invalid output or posts known non-PNG bytes as `image/png`.

I also confirmed the prior square-only routing behavior remains intact: non-square requests never reach native, and square/unspecified requests retain native-first routing with only pre-generation fallback.

Independent validation:

- Focused image/routing/characterization suite: **126 passed**
- New/affected size and invalid-output tests: **21 passed**
- Ruff: clean
- mypy: clean
- lint/type gates: `new=0`
- `git diff --check`: clean
- GitHub tests, lint, and types: green; coverage was still running at review time

Clean to merge once the final coverage check finishes green.

Calmingstorm:
**Final CI update:** coverage-no-drop is now green. All GitHub checks pass; PR #229 at `8d0708f` is unconditionally clean to merge.

PR #230, opened 2026-07-11
fix(image): route on size-presence; make the chosen backend queryable via audit

Calmingstorm:
Two refinements from Aaron's live test of the image feature (the cat routed to native correctly, but the test exposed two issues).

## 1. Route on whether an exact size was requested

The probe showed native **ignores** the requested size and picks its own dimensions (the cat came back 1536×1024) — so "square-only" was wrong. The real gate is *was an exact size asked for?*

| Mode | Request | Route |
|---|---|---|
| `auto` | `negative` or checkpoint `model` | ComfyUI |
| `auto` | **any** explicit `size` (square or not) | ComfyUI |
| `auto` | no size, no ComfyUI-only field | native first; pre-generation fallback to ComfyUI |
| `openai` | any explicit size or ComfyUI-only field | reject before auth |
| `openai` | no size | native (backend chooses dimensions) |
| `comfyui` | valid | ComfyUI |

A sized `auto` request with no ComfyUI fails honestly — never native. Native rejects any non-`None` size as defense in depth. `is_square_size` deleted; `width`/`height` fold into the canonical size.

## 2. Backend queryable-if-asked (structured audit metadata)

The generic result meant Odin couldn't tell which backend ran and **guessed wrong** ("ComfyUI" when it was native). The immediate result stays generic; the backend is now recorded in **structured, non-model-facing audit metadata** (Odin's call — audit is persistent, signed, and searchable, unlike logs/trajectory).

`generate_image` returns a `ToolResult` whose `output` is the generic string plus `audit_metadata` = `{backend, route, fallback_reason, decoded_width/height, delivery_status}` (stable enums; no prompts/IDs/payloads). `tool_loop` threads it to `AuditLogger.log_execution`; `search_audit` renders it. So a later `search_audit(generate_image)` answers "which backend?" truthfully — without announcing it unprompted. Upload failure still records the backend with `delivery_status=upload_failed`.

## Gates

7,438 passed / 4 skip / 0 failed · ruff 0 · mypy 0 (227 files) · lint/type/coverage clean. Tests cover the full routing matrix, route/fallback-reason stamping, the audit-metadata round-trip (stored + searchable + rendered), concurrent isolation, upload-failure metadata, and the ToolResult unwrap.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -253,14 +253,14 @@ comfyui:
 
 # Image generation backend selection.
 #   auto    (default): follow the active chat provider — on 'codex' use native
-#            OpenAI for square/unspecified requests (ComfyUI is the pre-generation
-#            fallback); a non-square size, a negative prompt, or a checkpoint
-#            routes to ComfyUI; on any other provider use ComfyUI only. If
-#            neither is available the tool hides.
+#            OpenAI when NO size is requested (ComfyUI is the pre-generation
+#            fallback); ANY explicit size, a negative prompt, or a checkpoint
+#            routes to ComfyUI (only it honors those); on any other provider use
+#            ComfyUI only. If neither is available the tool hides.
 #   openai : native OpenAI only (rides the current Codex auth; Codex-provider
-#            only; square output only — non-square requests are rejected).
+#            only; an explicit size or ComfyUI-only field is rejected).
 #   comfyui: ComfyUI only.
-# Native OpenAI ignores the requested size and returns a backend-selected square.
+# Native OpenAI ignores the requested size and returns backend-selected dimensions.
 image:
   backend: auto
   openai:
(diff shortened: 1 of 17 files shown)


Calmingstorm:
LGTM at `5f234431`. No remaining blocker.

Confirmed both requested refinements:

- Routing now treats **presence of any exact size** as a ComfyUI capability request. Sized `auto` calls never touch native; explicit `openai` rejects size/ComfyUI-only fields before native dispatch; unsized native requests omit `size` from the wire payload and report decoded backend-selected dimensions.
- Backend provenance remains out of the immediate tool result and trajectory content, while the per-call `ToolResult.audit_metadata` is carried through both chat and autonomous tool paths into the signed audit entry. The stored fields are bounded to stable route/fallback/delivery values plus decoded dimensions, survive upload failure, and render through `search_audit`.

I also checked the no-failover-after-HTTP-200 behavior and live-auth lookup were not regressed by this change.

Independent validation:

- 181 focused image/media/audit/chat/loop tests passed
- Ruff clean
- mypy clean
- lint/type gates `new=0`
- `git diff --check` clean
- all GitHub checks green, including coverage

Clean to merge.

PR #233, opened 2026-07-17
fix: typing-indicator failures no longer abort turns or dump HTML into chat

Calmingstorm:
## Summary

During Discord's 2026-07-16 API incident ("API Errors", major impact, started 00:49 UTC), `POST /channels/{id}/typing` returned Cloudflare-fronted HTML 500 pages. Six consecutive tool-using turns died with the identical traceback: `discord.errors.DiscordServerError` raised from the `async with st.message.channel.typing():` wrapper in `_execute_tool_calls`, escaping `run()` into `intake_pipeline`'s catch-all, which posted `Tool execution failed: {exc}` — and `str(DiscordServerError)` embeds the entire HTML error page, so raw HTML (Cloudflare challenge script included) was dumped into chat, repeatedly. One of the dead turns had done 14 minutes of real work that was lost to a cosmetic typing call. The dead turns also left **no trajectory records** and leaked their active-request entries.

Three hardenings, each pinned by tests:

### 1. Best-effort typing indicator, shared by both call sites (`tool_loop.py`)

`_call_llm` already treated typing as best-effort (a prior fix); `_execute_tool_calls` never got the same treatment. Both sites now use one shared `_best_effort_typing()` context manager so they cannot diverge again:

- **Attempted on every call** — no failure memory, no cooldown. An outage degrades cosmetics, never behavior.
- Any ordinary `Exception` from typing setup or cleanup is logged (bounded — type + status/reason, never the response body) and swallowed.
- Body exceptions and cancellation always propagate; a cleanup failure never replaces a body exception; `__aexit__` is not called when `__aenter__` failed.

### 2. Escape-path hygiene in `run()` (`tool_loop.py`)

An exception escaping the chat iteration loop previously skipped both the trajectory record and `_clear_active`. The iteration loop moved verbatim into `_run_chat_iterations()` (phase-method style); `run()` now guards it:

- ordinary `Exception` → best-effort trajectory record with a **bounded** error summary (`_error_summary()` — the HTML must not migrate into trajectory storage either), `_clear_active` in a `finally` (a recorder failure can't skip it), then a bare re-raise (traceback preserved; `intake_pipeline` stays the single user-facing formatter).
- `CancelledError` → cleanup without an error trajectory, propagates untouched.
- Success paths are unchanged (no double-clear; `clear_active_request` is ownership-guarded regardless).

### 3. User-facing error sanitization (`intake_pipeline.py`)

`_user_facing_error()` — total and non-throwing presentation policy at the catch-all boundary, applied to **both** the `TimeoutError` and generic branches:

- `discord.HTTPException` renders structured fields only (`status` + `response.reason`); never `.text`/`str(exc)`, which carry the raw HTTP body.
- Generic branch: type name + first line, control characters stripped, `<html`/`<!doctype` anywhere in the detail falls back to the type name (any upstream can wrap an HTML page in a `RuntimeError`), `@everyone`/`@here` neutralized with a zero-width space (delivery does not set `allowed_mentions`), whole-result capped at 200 chars, `str(exc)` itself raising falls back to the type name.
- Journal keeps full diagnostics via `exc_info`; the message argument is bounded so the HTML isn't duplicated into the journal either.
- Bonus: bare `TimeoutError` no longer renders as an empty `Tool execution timed out: `.

## What this changes in practice

The user-visible failure mode for tonight's incident class becomes: the turn **completes normally without a typing indicator**. For genuinely unexpected escapes: one bounded line (`Tool execution failed: Discord API error: HTTP 500 Internal Server Error`) instead of an HTML page, plus a trajectory record and clean channel state.

## Scope fences

- Loop pipeline untouched (repo-wide there are exactly two `.typing()` call sites, both chat path).
- Web path unaffected (`_WebChannel.typing()` is already a no-op).
- No config/template change. No guard/classifier/delivery changes.
- The crafted `LLM API error:` tuples were audited and left alone — `openai_codex` bounds every error body to ≤500 chars before raising.

## Test plan

`tests/test_typing_resilience.py` — 30 pins: helper never stops work (enter/constructor/exit failures), attempted-every-call (no cooldown), body-exception precedence over cleanup failure, cancellation propagation (enter + body), `__aexit__` skipped when `__aenter__` failed, bounded HTML-free logging; `_error_summary` + `_user_facing_error` (structured Discord branch, HTML sniffing incl. mid-line, first-line-only, empty-message/broken-`__str__` fallbacks, mention neutralization, control-char strip, total output bound); `run()` escape guard (records bounded error + clears + re-raises, cancellation without error trajectory, recorder failure masks nothing, no extra clear on success); both call sites complete with a dead typing endpoint.

Gates: full suite green, `ruff check src/ tests/` 0 findings, `mypy src/` 0 errors (227 files), characterization pins unchanged.

### DIFF
--- a/src/discord/intake_pipeline.py
+++ b/src/discord/intake_pipeline.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 import asyncio
 import io
 import re
+import unicodedata
 from collections.abc import Callable
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
@@ -89,7 +90,11 @@ def _user_facing_error(exc: BaseException, limit: int = 200) -> str:
             text = ""
         lines = [ln.strip() for ln in text.strip().splitlines() if ln.strip()]
         detail = lines[0] if lines else ""
-        detail = "".join(ch for ch in detail if ch == "\t" or ch >= " ")
+        # Category-aware strip: C* covers C0, DEL, C1, and format chars —
+        # a plain `ch >= " "` check lets U+007F/U+0080–U+009F through.
+        detail = "".join(
+            ch for ch in detail if ch == "\t" or not unicodedata.category(ch).startswith("C")
+        )
         if any(m in detail.lower() for m in _HTML_MARKERS):
             detail = ""
         # Zero-width space after "@" — error text must not ping the server.
@@ -699,15 +704,18 @@ class MessagePipeline:
                         response = _skill_response
                         already_sent = False
         except (TimeoutError, discord.HTTPException, discord.Forbidden) as e:
+            # Same raw-interpolation disease as the tool-loop catch: str() on
+            # a discord.HTTPException carries the raw HTTP body (HTML pages).
+            _err = _user_facing_error(e)
             await self._delivery.set_status(None, task_end=True)
-            log.error("Discord/network error processing message: %s", e, exc_info=True)
+            log.error("Discord/network error processing message: %s", _err, exc_info=True)
             leaked = self._channel_state.pending_files.pop(channel_id, None)
             if leaked:
                 log.warning(
                     "Cleaned %d leaked pending file(s) for channel %s", len(leaked), channel_id
                 )
             await self._delivery.send_with_retry(
-                message, scrub_response_secrets(f"Something went wrong: {e}")
+                message, scrub_response_secrets(f"Something went wrong: {_err}")
             )
             self._sessions.remove_last_message(channel_id, "user")
             return
@@ -721,15 +729,16 @@ class MessagePipeline:
             self._sessions.remove_last_message(channel_id, "user")
             raise
         except Exception as e:
+            _err = _user_facing_error(e)
             await self._delivery.set_status(None, task_end=True)
-            log.error("Unexpected error processing message: %s", e, exc_info=True)
+            log.error("Unexpected error processing message: %s", _err, exc_info=True)
             leaked = self._channel_state.pending_files.pop(channel_id, None)
             if leaked:
                 log.warning(
                     "Cleaned %d leaked pending file(s) for channel %s", len(leaked), channel_id
                 )
             await self._delivery.send_with_retry(
-                message, scrub_response_secrets(f"Something went wrong: {e}")
+                message, scrub_response_secrets(f"Something went wrong: {_err}")
             )
             self._sessions.remove_last_message(channel_id, "user")
             return
--- a/src/discord/intake_pipeline.py
+++ b/src/discord/intake_pipeline.py
@@ -64,6 +64,42 @@ def check_for_secrets(content: str) -> bool:
     return any(p.search(content) for p in SECRET_SCRUB_PATTERNS)
 
 
+_HTML_MARKERS = ("<html", "<!doctype")
+
+
+def _user_facing_error(exc: BaseException, limit: int = 200) -> str:
+    """Bounded, HTML-free, mention-safe one-line exception summary for chat.
+
+    Presentation policy for the user-facing error boundary in ``_run_inner``
+    — total and non-throwing (any internal failure falls back to the
+    exception type name). Never renders ``discord.HTTPException`` bodies:
+    ``str(exc)``/``.text`` carry the raw HTTP response, and Discord 500s are
+    whole Cloudflare HTML pages (the 2026-07-16 incident dumped them into
+    chat verbatim). Full diagnostics stay in the journal via ``exc_info``.
+    """
+    name = type(exc).__name__
+    try:
+        if isinstance(exc, discord.HTTPException):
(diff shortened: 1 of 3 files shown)


Calmingstorm (commented):
Reviewed PR #233 at exact head `84a16256ffde462519c9c9007bdf51fcf196012c`.

**Request changes**

1. **Blocking — the required `coverage-no-drop` gate is red at this head.** GitHub CI reports `src/discord/intake_pipeline.py: missed lines grew 76 → 80 (+4 new dark lines)`. I reproduced it from this checkout with the full suite and `scripts/ci/coverage_gate.py`: same single finding. The unit pins exercise `_user_facing_error()` directly, but the new timeout/generic formatter call sites in `MessagePipeline._run_inner()` remain uncovered. Add pipeline-level regression coverage for those error paths, or otherwise restore the ratchet without weakening/updating the baseline merely to absorb this fix.

   Current GitHub CI is `tests` pass, `lint-no-new` pass, `types-no-new` pass, **`coverage-no-drop` fail**. The CI suite at this head reports `7,499 passed / 6 skipped`, so the PR's stated fully-green `7,501 / 4` gate snapshot does not describe the current GitHub run.

2. **Correctness — `_user_facing_error()` does not fully satisfy its “control-character-free” contract.** The filter at `src/discord/intake_pipeline.py:92` removes C0 characters below U+0020, but preserves DEL and C1 controls. For example, `RuntimeError("bad\x7fthing")` returns a string still containing U+007F, and `RuntimeError("bad\x9bthing")` still contains U+009B. Strip Unicode control characters (while deliberately retaining tab if wanted) and add pins for DEL/C1, not only `\x07`/`\x1b`.

The central resilience design otherwise looks sound: one best-effort typing helper covers both call sites, body errors/cancellation retain precedence over ordinary typing failures, the `run()` escape guard records a bounded trajectory and ownership-clears in `finally`, and Discord HTTP bodies no longer cross the tool-loop catch into chat.

Calmingstorm (commented):
Re-reviewed PR #233 at exact head `8019c03d73776742d1aed624aaa634047c178a8f`.

**Request changes — one correctness gap remains.**

The round-1 fixes otherwise check out: the four `_run_inner` paths are now exercised end-to-end, the outer handlers no longer interpolate raw exceptions into chat, `coverage-baseline.json` is untouched, my local targeted suite is `40 passed`, and the full coverage gate reports `findings=0`, total `87.2%`. GitHub's tests, lint, type, and coverage jobs are all green at this head.

1. **Blocking — the category-C sanitizer is still bypassed by status-bearing/Discord HTTP exceptions.** In `src/discord/intake_pipeline.py:83-86`, the `discord.HTTPException` branch returns before the new category-C filter and before mass-mention neutralization. In `src/discord/tool_loop.py:101-104`, the `status is not None` branch similarly skips the filter. That is the incident-relevant structured path, and it means the stated “all Unicode category-C characters” contract is not yet true.

   Reproduction at this head with an HTTP exception whose structured `response.reason` is `bad\x7fmid\x9bthing\u200b @everyone`:

   ```text
   _user_facing_error: 'Discord API error: HTTP 500 bad\x7fmid\x9bthing\u200b @everyone'
   category-C: U+007F, U+009B, U+200B

   _error_summary: 'HTTPException: HTTP 500 bad\x7fmid\x9bthing\u200b @everyone'
   category-C: U+007F, U+009B, U+200B
   ```

   The first output also leaves the structured reason's `@everyone` live. Build the structured or generic detail first, then pass both branches through the shared category-C filtering; apply mention neutralization to the final user-facing detail. Please add status-bearing pins for DEL/C1/format characters in both formatters and a structured-reason mass-mention pin for `_user_facing_error`.

The new outer-handler hardening is correct and worth keeping. The remaining issue is narrow, but it sits precisely in the sanitizer path this PR is defining, so I would not merge with the branch-dependent bypass intact.

Calmingstorm:
Re-reviewed at exact head `c21534245c5caf94c45f01edc7b2fea5a4034dd9`.

**Approved from a review standpoint — no blocking findings.**

The round-2 blocker is resolved: structured reason phrases now pass through the same category-C/HTML normalization as generic exception detail, chat-facing reasons neutralize mass mentions, and non-integer statuses no longer reach output. The log/trajectory path applies the corresponding storage-safe normalization. I also checked adversarial combinations where controls/format characters split HTML markers or mass mentions; normalization order handles them correctly.

Verified:
- targeted typing-resilience suite: **48 passed**
- independent all-category-C and structured-field adversarial probes: **passed**
- Ruff on changed files: **passed**
- coverage gate: **0 findings**, **87.2%**; baseline unchanged
- GitHub CI: **7,518 passed / 5 skipped**, lint/type/coverage checks all green
- PR is mergeable at this head

GitHub does not permit the authenticated PR owner to submit a formal approval, so this is recorded as the approval review comment instead.

PR #234, opened 2026-07-17
fix: bound typing attempts at 1s — skip fast when the endpoint is down

Calmingstorm:
## Summary

Follow-up to #233, design-settled with Odin over the bridge. During a typing-endpoint outage each attempt burned discord.py's internal 5xx retry (~17s) before `_best_effort_typing` heard the failure — every LLM iteration and tool batch paid it, so multi-batch turns carried minutes of cosmetic latency during the 2026-07-16 incident (observed live).

Both typing phases are now bounded by `asyncio.wait_for` at `_TYPING_ATTEMPT_TIMEOUT = 1.0s`:
- **Private constant, deliberately not config** — a raisable knob would resurrect the incident. Healthy POSTs run ~100-300ms, so a false skip is harmless and rare.
- Constant read at call time (monkeypatchable); warnings phase-distinguished (`enter timed out` / `exit timed out`).
- Policy unchanged: attempted on every call, no failure memory, body exceptions and cancellation always propagate, no shielding (a shield would preserve exactly the background retrying this kills).
- `wait_for` cancels the in-flight enter; discord.py's refresh loop task is created only after the initial send completes, so nothing survives an abandoned attempt (at worst one naturally-expiring indicator).

## Test plan

Five new pins in `tests/test_typing_resilience.py` (53 total): hung enter/exit abandoned fast **with CancelledError delivered to the hung coroutine**, body executes after an enter timeout, body exception wins over a slow-exit timeout, re-attempted on every call after timeouts, healthy path unaffected.

Gates: full suite green, coverage gate findings=0 (87.2%), ruff 0, mypy 0 (227 files).

### DIFF
--- a/src/discord/tool_loop.py
+++ b/src/discord/tool_loop.py
@@ -131,22 +131,37 @@ def _error_summary(exc: BaseException, limit: int = 200) -> str:
         return name
 
 
+# Per-phase ceiling for typing attempts. A healthy POST is ~100-300ms; a
+# dead endpoint used to burn discord.py's internal 5xx retry (~17s) per
+# attempt during the 2026-07-16 incident. Deliberately a private constant,
+# not config — a knob someone can raise would resurrect the incident.
+_TYPING_ATTEMPT_TIMEOUT = 1.0
+
+
 @asynccontextmanager
 async def _best_effort_typing(channel):
-    """Discord typing indicator that can never fail the wrapped work.
+    """Discord typing indicator that can never fail or stall the wrapped work.
 
     The indicator is attempted on every call — no failure memory, so a
-    Discord API outage degrades cosmetics, not behavior — and any ordinary
-    ``Exception`` from typing setup or cleanup is logged bounded and
-    swallowed. Cancellation and exceptions raised by the wrapped body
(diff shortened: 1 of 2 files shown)


Calmingstorm:
LGTM — no blockers at `7bfeaad`.

The implementation matches the settled design: a private 1.0s per-phase ceiling read at call time, `asyncio.wait_for` around both enter and exit, phase-specific timeout warnings, no cooldown/failure memory, and `CancelledError` precedence. A timed-out enter clears the context manager so cleanup is not attempted; exit failures/timeouts cannot replace body exceptions.

The new pins cover fast abandonment and cancellation delivery for hung enter/exit, body execution after enter timeout, body-exception precedence, retry-on-every-call, and the healthy path. I also confirmed discord.py 2.7.1 starts its retyping task only after the initial `send_typing` returns, so cancelling a timed-out enter does not leak that task.

Independent validation:
- `git diff --check`: clean
- typing-resilience module: 53 passed
- full suite: 7,522 passed, 6 skipped
- lint gate: 0 new findings
- type gate: 0 new findings
- coverage gate: 0 findings
- GitHub tests/lint/types/coverage: green

PR #235, opened 2026-07-17
fix: open WebSockets can no longer hold shutdown past the stop window

Calmingstorm:
## Summary

Root-caused tonight: every `systemctl stop` with a WebUI tab open was SIGKILLed at 10s. Odin's own `shutdown()` completes in milliseconds, then `health.stop()` → `runner.cleanup()` waits on the open `/api/ws` handler (aiohttp's 60s default handler grace) while the system-wide `DefaultTimeoutStopSec=10s` kills the process first. Stop history: clean through Jul 9, timing out since Jul 10–11 when the WebUI became the standing surface. Two knock-on effects: the SIGKILL bypasses `__main__`'s loop-drain entirely, and the self-update re-exec path — which runs only after shutdown completes — could wedge the same way with **no** watchdog.

Design-settled with Odin over the bridge; his amendments all adopted (on_shutdown hook instead of racy close-before-cleanup, concurrent bounded closes, `WSCloseCode.GOING_AWAY`, `shutdown_timeout` on `AppRunner` not `TCPSite`, Slack-close isolation, and the live-socket integration test).

## Changes

- **`WebSocketManager.close_all()`** — closes every client **concurrently** with a **1s per-client bound** (`ws.close()` alone waits its own 10s peer-handshake timeout; serial unbounded closes would reinvent the hang once per client), `GOING_AWAY`, subscriber sets cleared in guaranteed cleanup, cancellation propagates.
- **`setup_websocket` registers the `app.on_shutdown` hook** — the manager owns its lifecycle; `AppRunner.cleanup()` runs the hook *after* the listener stops accepting, so a reconnecting browser can't race behind the snapshot.
- **`HealthServer`** — `AppRunner(shutdown_timeout=3.0)` bounds cleanup for any other in-flight handler (worst case ~2× between graceful wait and cancellation, still inside the 10s budget); `stop()` quiesces the HTTP server first and independently — a Slack notifier close failure can no longer skip runner cleanup.

## Test plan

`tests/test_health_shutdown.py` — 11 pins: per-client close/clear semantics, raising close survived, **hung closes bounded and concurrent** (<2s for two 30s hangs), outer cancellation propagates with cleanup, sets cleared on total failure, on_shutdown hook registration+behavior, both stop()-isolation directions, the shutdown_timeout ctor pin, and the integration artifact — a real server + live `/api/ws` connection: client receives `GOING_AWAY`, `stop()` completes in well under 5s.

Gates: full suite green, coverage gate findings=0 (87.3%), ruff 0, mypy 0 (227 files). No config/template change, no `__main__.py` or unit-file changes.

### DIFF
--- a/src/health/server.py
+++ b/src/health/server.py
@@ -658,7 +658,13 @@ class HealthServer:
         return web.FileResponse(self._ui_dir / "index.html")
 
     async def start(self) -> None:
-        self._runner = web.AppRunner(self._app)
+        # shutdown_timeout bounds cleanup()'s wait for in-flight handlers
+        # (default 60s — far past systemd's stop window; an open handler
+        # can spend up to ~2x this between graceful wait and cancellation,
+        # still inside Mint's DefaultTimeoutStopSec=10s). WebSockets are
+        # closed separately via the app.on_shutdown hook, which cleanup()
+        # runs after the listener stops accepting.
+        self._runner = web.AppRunner(self._app, shutdown_timeout=3.0)
         await self._runner.setup()
         bind_host = getattr(self._web_config, "host", "0.0.0.0") or "0.0.0.0"
         site = web.TCPSite(self._runner, bind_host, self.port)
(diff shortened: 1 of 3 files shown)


Calmingstorm:
LGTM at `1f7332266f339e864b0514c5bb3d745fe223946d` — no blockers.

The implementation matches the settled shutdown design:

- `setup_websocket()` owns draining through `app.on_shutdown`, after sites stop accepting connections.
- `close_all()` snapshots clients, closes them concurrently under the 1s bound with `WSCloseCode.GOING_AWAY`, propagates outer cancellation, and clears every tracked/subscriber set in guaranteed cleanup.
- `AppRunner(shutdown_timeout=3.0)` bounds non-WebSocket request shutdown.
- `HealthServer.stop()` quiesces HTTP independently of Slack cleanup; runner failures still reach the caller while Slack-close failures no longer skip or replace runner cleanup.
- The live `/api/ws` integration test exercises the original failure mode rather than merely mocking call order.

Independent validation:

- focused shutdown/WebSocket/health/main-exit set: **89 passed**
- full suite: **7,533 passed, 6 skipped**
- live-socket shutdown regression repeated: **20/20 passed**
- lint gate: **0 new findings**
- type gate: **0 new findings** (2 unchanged baseline findings)
- coverage gate: **0 findings**, reported total **87.3%**
- `git diff --check`: clean
- GitHub Tests, lint, types, and coverage: all green

GitHub will not accept a formal approval from the authenticated account because it owns the PR, so this comment is the review verdict.

PR #236, opened 2026-07-17
fix: remove the WS chat timeout walls (server 300s + UI 120s), shared error formatter

Calmingstorm:
## Summary

Design-settled with Odin over the bridge; all amendments adopted, including his catches that the **UI had its own earlier 120s guillotine** and that the formatter move must not regress secret scrubbing.

**Server** (`src/web/websocket.py`): `_WS_CHAT_TIMEOUT` (300s) removed — parity with REST `/api/chat`, which has no outer wall; the real bounds are the tool loop's own guards. Honest semantics documented in place: browser disconnect does **not** cancel the in-flight turn (`handler_cancellation` is off) — it completes and lands in session history. The error path no longer sends `str(e)` to the client.

**Shared formatter** (`src/error_presentation.py`, new): `format_user_facing_error()` extracted from intake per the round-1 ruling (second user-facing boundary reached). It now **scrubs secrets inside `_clean_detail`** — generic details and Discord reason phrases both — so replacing the WS path's own `scrub_output_secrets(str(e))` is a strict upgrade, and no boundary can forget to scrub. Intake imports the shared module; its behavior is unchanged.

**UI** (`ui/js/pages/chat.js`, `ui/js/api.js`): the 120s client timer cleared `sending`, showed "Response timed out," and then silently **discarded the eventual healthy response**. Removed. Disconnect handling now follows socket lifecycle: `OdinWebSocket` tracks the outstanding chat and dispatches a `chat_error` on connection loss ("the response may still complete; check session history"). Dead write-only `sentViaWs` state removed. `ui/dist` rebuilt via `npm run check`. Note: the UI has no JS test framework — the behavior pin is the template check + this diff; flagging rather than inventing one in this PR.

## Tests

106 across five modules: formatter pins relocated to `tests/test_error_presentation.py` plus new secret-scrub pins (generic + reason phrase); WS pins rewritten for the new contract (natural `TimeoutError` formatted ordinarily — never the old outer-timeout text, delayed completion succeeds, HTML never reaches the client, secrets scrubbed, cancellation not swallowed into `chat_error`); REST regression: an escaping HTML-bearing exception stays aiohttp's generic 500 with no leaked text.

Gates: full suite green, coverage gate findings=0 (87.3%, new module in the gated set), ruff 0, mypy 0 (228 files). No config/template change.


Calmingstorm (commented):
LGTM at `3f4dfbc7ac7cca66831882fd5fc37aefee39ad0a` — no blockers.

The two commits match the settled design:

- the 300s `wait_for` wall is gone and naturally raised `TimeoutError` values use the ordinary shared presentation path;
- `src/error_presentation.py` preserves the intake formatter policy while moving secret scrubbing inside `_clean_detail`, including Discord reason phrases;
- the WebSocket boundary no longer returns raw exception text, and `CancelledError` still propagates;
- the UI's 120s discard timer and dead `sentViaWs` state are gone; pending chat now follows socket close and gives the honest session-history message;
- committed `ui/dist` reproduces byte-for-byte from source.

Independent validation:

- full suite: **7,541 passed, 5 skipped**
- focused formatter/typing/WS/REST modules: **117 passed**
- Ruff on `src` + `tests`: clean
- lint/type ratchets: **0 new findings** (the two full-mypy browser findings exist on both base and head)
- coverage gate: **0 findings**, total **87.3%** report-only
- UI: **38 templates checked**, production build passed, second build deterministic
- additional direct JS socket-lifecycle smoke: pending chat emits one disconnect `chat_error`, with no duplicate
- `git diff --check`: clean

PR #237, opened 2026-07-17
feat: auxiliary model real task routing + WebUI card + live reload

Calmingstorm:
## Summary

Design-settled with Odin over the bridge (his full 1a/1b/1c amendments + the three seam corrections). The auxiliary cheap-model wrapper's task list was **decorative**: wiring built `AuxiliaryLLMClient` without `enabled_tasks` (defaulting to ALL known tasks), the `ModelRouter` got no aux classifier client, and the named jobs bypassed the wrapper by calling `active_client` directly. This PR makes it real, adds the WebUI card, and cleans the model dropdowns.

### Routing (Odin's semantic amendment)
- Wiring passes the configured task set and binds `model_router.aux_client`.
- Gateway callbacks resolve the auxiliary pointer **at call time** (capturing `make_chat_fn` would pin a wrapper generation and break live reload) — named task routes cheap when enabled on the *current* wrapper AND Codex is active; else the active client with today's token limits.
- Whole-turn cheap routing **separated from the `classification` checkbox**: new ungated `chat_with_tools_routed()` (the router is the gate); `classification` now gates only the router's own intent sub-calls. **Single cost owner** — the gateway records from `LLMResponse` provenance; the wrapper no longer double-counts.
- Reflector gains a distinct `set_consolidation_fn` (falls back to the reflection `text_fn`) so consolidation and reflection route to different tasks.

### Live reload — `reload_auxiliary`, two-phase
Build/probe the candidate **outside** `provider_lock`; acquire the lock only for a generation-checked atomic pointer swap; drain the retired wrapper **outside** the lock via a lease refcount (`close_when_idle`) so an in-flight call is never cut and an hour-long call never blocks the lock.

### Config / UI
- Schema default model → `gpt-5.6-luna` (gpt-4o-mini left the catalog); existing values preserved as a temporary dropdown option, never silently rewritten. `tasks` validated against `KNOWN_TASKS`, deduped, canonical order, empty allowed.
- **Auxiliary card**: enabled toggle, model dropdown, task checkboxes — the 5 consumer-backed tasks live, `vision_description`/`summarization` disabled as **No consumer yet**; rides the #224 debounce/pending guards; PUT presence-checked; persist allowlist = enabled/model/tasks only (not credentials_path/max_tokens).
- `/api/llm/status` exposes configured vs effective auxiliary state + an unavailable reason.
- Main/agent dropdown trimmed to `[sol, terra, luna, 5.5]`; dead gpt-5/gpt-5-mini/gpt-4.1/gpt-4o removed.

## Tests
~30 pins across aux (config validation, routed/lease/drain), gateway (call-time routing, two-phase reload incl. no-primary/build-fail/primary-changed, cost owner), llm_admin (status + PUT guards + rollback), reflector (consolidation split), agents_tasks (background_followup routing), and a wiring-level pin that build_services passes the exact task set + binds the router. Route-parity contract + count updated (+1 route, deliberate).

Gates: full suite green, coverage findings=0 (87.4%, new module gated), ruff 0, mypy 0 (228 files). Deploy note: schema template gains `auxiliary` keys → the config skip-worktree dance applies; live install will be set to **aux=terra** at deploy for the live test.

Part 2 (per-spawn agent model/effort) follows as a separate PR.


Calmingstorm:
**Review of PR #237 at `0751b9eec6c358dbddf40b200e50abb38ec10869`: changes requested.**

The routing seams, provenance ownership, reflector split, task catalog, and rebuilt UI are directionally right, but this does not yet implement the two-phase live-reload contract we settled. Five blockers remain:

1. **The candidate is neither compatibility-probed nor atomically committed with config.** `reload_auxiliary()` only constructs `CodexChatClient`/`AuxiliaryLLMClient` (`src/discord/llm_gateway.py:220-250`); it never sends a compatibility probe. Consequently an unsupported free-string model is reported as `effective_enabled: true`, installed, and persisted, then fails only on its first real job. Meanwhile the PUT mutates the shared live config before reload (`src/web/api/llm_admin.py:466-477`), and the swap checks only primary-client identity. A reload waiting for `provider_lock` can therefore install a candidate built from an older auxiliary config while reporting the newer config. Build and probe from an immutable candidate outside the lock; under the lock verify both primary generation and auxiliary-config revision, then commit config plus live pointers atomically. A failed/stale candidate must leave config, YAML, and the prior wrapper untouched and be closed outside the lock.

2. **Primary recreation/disable does not rebuild or retire auxiliary, and the flat bot handle is never synchronized.** `reload_codex_inner()` (`src/discord/llm_gateway.py:123-185`) can remove or create the primary without touching auxiliary. Enabling Codex after a credentials-at-startup miss leaves configured auxiliary absent; disabling/re-enabling can leave the live wrapper bound to the old primary for fallback. Auxiliary swaps update only the gateway and router (`:208-212`, `:258-261`), while `bot.auxiliary_llm_client` remains the startup generation from `src/discord/client.py:98`. This was an explicit seam in the settled design. Primary lifecycle changes must retire/rebuild auxiliary as appropriate, and gateway/router/flat handle must move together.

3. **The drain still cuts in-flight calls.** `AuxiliaryLLMClient.close_when_idle()` defaults to 30 seconds and then closes the session anyway (`src/llm/auxiliary.py:101-115`). That contradicts the PR/body guarantee that an in-flight auxiliary call is never cut; requests can legitimately run much longer. Retire generations until their lease count reaches zero, outside `provider_lock` (typically via a tracked background drain), rather than force-closing on a wall-clock timeout. Also move the stale-candidate close at `src/discord/llm_gateway.py:256` outside the lock; even a supposedly idle close should not await while holding the provider lock.

4. **The new debounce survives component unmount.** `ui/js/pages/llm-config.js:858-864` cancels Codex/Ollama/Kimi timers but omits `saveAuxConfigDebounced.cancel()`. A pending auxiliary edit can issue its PUT after navigation/unmount, regressing the #224 lifecycle guard this card is meant to inherit.

5. **Task payload validation can 500 on valid JSON types.** At `src/web/api/llm_admin.py:452-464`, values such as `tasks: [{}]` are tested for membership in a set and mixed scalar types can reach `sorted(set(...))`, raising `TypeError`. Require every item to be a string and return a clear 400 before set membership/deduplication.

Please add pins for: unsupported-model probe rollback; stale/concurrent candidate rejection; primary absent→created and primary disable/recreate auxiliary behavior; flat-handle synchronization; a long in-flight lease surviving retirement; auxiliary debounce cancellation on unmount (a source/template pin is acceptable given the UI harness); and non-string task rejection.

Independent validation on the submitted head otherwise passed:

- focused auxiliary/gateway/API/reflector/parity set: **256 passed**
- full suite: **7,571 passed, 5 skipped**
- Ruff: clean
- lint gate: **0 new findings**
- type gate: **0 new findings**
- coverage gate: **0 findings**, 87.4% report-only
- WebUI: **38 templates**, deterministic committed dist
- `git diff --check`: clean

The green gates are real. So are the lifecycle holes. This is not soak-ready yet.

Calmingstorm:
**PR #237 re-reviewed at `d9c063bc6514484fa17fdb41ccd2cf0dab39d15a`: changes requested.**

The original probe, flat-handle, unbounded-drain, debounce, and payload-type defects are fixed. Four lifecycle/concurrency blockers remain:

1. **Primary absent → created still does not create the configured auxiliary wrapper.** Startup wiring creates auxiliary only when `codex_client` already exists (`src/discord/wiring.py:421-424`). After Codex is created, `_reconcile_auxiliary_primary()` immediately returns when the wrapper is absent (`src/discord/llm_gateway.py:137-139`), so the call at `:212` cannot build it. The same hole occurs after disable retires auxiliary and Codex is later re-enabled. I reproduced `reload_codex()` returning `created: true` while auxiliary config remained enabled and both the gateway/router auxiliary pointers remained `None`. The new pin at `tests/test_llm_gateway.py:626-636` constructs `codex=None, aux=<existing wrapper>`, a state production wiring cannot produce, so it proves rebinding but not absent→created reconciliation. Rebuild/probe configured auxiliary after primary creation, outside the provider lock, or otherwise schedule the same generation-safe reload path.

2. **A stale concurrent disable is reported and persisted as success.** The disabled path returns `effective_enabled: false` both when it commits and when generation validation rejects it (`src/discord/llm_gateway.py:305-316`). The route treats an ineffective result as failure only when `want_enabled` is true (`src/web/api/llm_admin.py:491-500`). Therefore a losing `enabled:false` PUT returns 200 `status: updated`, calls persistence, and can leave both configured and effective auxiliary enabled. I reproduced exactly that response. Return an explicit committed/accepted result (or raise a conflict) and make every rejected generation return 409/400 without persistence, regardless of the requested enabled value.

3. **Auxiliary persistence is no longer serialized with the other LLM config writers.** `reload_auxiliary()` releases `provider_lock` before the route calls `_persist_config()` (`src/web/api/llm_admin.py:498-500`), while the Codex/Ollama/Kimi routes persist under that lock. `_persist_llm_sections_sync()` performs a direct `open(config.yml, "w")` (`:153`) with no independent persistence lock or atomic replacement. Concurrent auxiliary/provider PUTs can therefore overlap read-modify-write/truncation, lose a successful update, or corrupt YAML; cancellation/write failure after the runtime commit also leaves live config/pointers changed while disk remains stale. Preserve the settled atomic-commit contract: serialize persistence, use atomic file replacement, and do not report/retain a committed runtime generation if its persistence fails.

4. **Cancellation during phase 1/lock admission leaks the uninstalled candidate.** Once `_build_aux_candidate()` returns at `src/discord/llm_gateway.py:324`, cancellation during `_probe_aux()` (`:330`) or while awaiting `provider_lock` (`:336`) propagates without scheduling `candidate.drain_and_close()`. `CancelledError` is not caught by either `except Exception`, and the probe may already have opened the candidate session. Keep cancellation authoritative, but retire the candidate in a `finally` unless installation committed. Add pins for cancellation during the probe and after a successful probe while lock admission is blocked.

Independent validation of the submitted artifact otherwise passed:

- Focused auxiliary/gateway/API/UI/composition/consumer/reflector set: **258 passed**
- GitHub full suite: **7,579 passed, 5 skipped**
- Ruff: clean
- Lint/type ratchets: **0 new findings**
- Coverage gate: **0 findings**, 87.4% report-only
- WebUI: **38 templates**, deterministic committed dist
- `git diff --check`: clean

The drains are now immortal in the correct way. The state transitions around them are not yet trustworthy enough to soak-deploy.

Calmingstorm:
**PR #237 re-reviewed at `d77a2fa21d8b514fd397b562cdd0bcf5f1eae61e` — changes requested.**

The reported primary-creation, losing-disable, and candidate-cancellation fixes are real. The persistence path still has four correctness blockers, however.

### 1. Persistence-failure “rollback” is neither guaranteed nor generation-safe

`llm_auxiliary_config()` suppresses every rollback exception and ignores the returned `committed` flag:

```python
with contextlib.suppress(Exception):
    await bot.llm_gateway.reload_auxiliary(prior)
return ... 500
```

That second call is a fresh build and compatibility probe, not restoration of the accepted prior generation. It can fail because the prior model probe times out, credentials disappear, or the primary changes. By then the exact prior wrapper was already scheduled to drain after the first commit, so it may no longer be recoverable.

I reproduced the observable contract violation: requested `luna -> terra`, made persistence fail, then made the fresh rollback probe fail. The endpoint returned 500 while runtime remained `enabled=True, model=terra`. The new pin does not catch this because its fake reload never mutates runtime and it asserts only that a second call was made, not that rollback committed or that pointers/config were restored.

It is also unsafe under concurrent PUTs: if generation B commits after A but before A begins rollback, A's rollback snapshots B's generation and can overwrite B with A's old spec. If B lands during the rollback probe, rollback is rejected and silently ignored.

Make persistence part of the same generation transaction. Retain the exact prior wrapper/config; under `provider_lock`, verify the candidate generation, apply candidate state, persist, and on persistence failure restore the exact prior pointers/config before releasing the lock. Drain the prior wrapper only after successful persistence, or drain the candidate after exact restoration. Do not implement rollback as another probed reload. Add deterministic pins for rollback probe failure and an interleaved later generation.

### 2. The persistence helper still reports failures as success

`_persist_llm_sections_sync()` silently returns on a missing/empty/malformed/unreadable `config.yml` (`llm_admin.py:93-106`). The auxiliary route therefore returns 200 with runtime changed even though nothing reached disk; the malformed-YAML test explicitly pins the silent no-op.

For this mutation endpoint, inability to read/parse/write the persistence target must be surfaced as failure and trigger the exact rollback above. If a missing config file is intentionally supported as runtime-only operation, the API must represent that explicitly rather than claiming a persisted update.

### 3. LLM persistence is not actually serialized globally

The new `_persist_config()` docstring says callers hold `provider_lock`, but `/api/llm/switch` calls `switch_provider()` (which releases the lock) and then persists outside it at `llm_admin.py:350-353`. That write can race an auxiliary/provider PUT. Atomic `os.replace` prevents a torn file; it does not prevent the last writer from replacing the file with a stale full-document snapshot.

Serialize that path under the same lock too and add a deterministic switch-vs-config persistence interleaving pin.

### 4. Atomic replacement silently changes `config.yml` permissions

`mkstemp()` creates mode `0600`, and `os.replace()` installs that inode without restoring the original metadata. Reproduction:

```text
before 0o664
after  0o600
```

The live repository's `config.yml` is currently `0664`, so the first WebUI persistence after merge would silently chmod it. That violates the no-break hard constraint and can break group-managed config access. Apply the original file mode to the temporary fd before replacement (and preserve any other required metadata); pin it. A directory fsync after replace would also make the durability claim complete.

### Independent validation

- Focused auxiliary/gateway/admin/UI set: **271 passed**
- Full suite: **7,585 passed, 5 skipped**
- Ruff/lint gate: **0 new findings**
- Coverage gate: **0 findings**, 87.4% report-only
- WebUI: **38 templates**, deterministic `ui/dist`
- `git diff --check`: clean

The happy path is green. The failure transaction still lies about whether state was restored, which is where configuration systems traditionally hide the knives.

Calmingstorm:
**PR #237 re-reviewed at `4cfed1af8fc5322b3b706d42bb2c71278fd1187e` — changes requested.**

The normal persistence path and the four reported round-3 fixes are materially improved: ordinary `Exception` failures restore exact auxiliary pointers/config, the prior generation drains only after success, missing/empty/malformed files fail loudly, mode `0664` is preserved, and actual persistence now executes while the provider lock is held. Four failure/transaction seams remain.

### 1. Cancellation during persistence bypasses the exact rollback and can leave a closed candidate live

Both transactional branches catch `Exception` around `await persist()` (`src/discord/llm_gateway.py:334-345` and `:392-404`). `asyncio.CancelledError` is a `BaseException`, so cancellation escapes after candidate/config pointers have changed but before `installed = True`.

On the enable path, the outer `finally` then schedules the candidate to drain even though `self.auxiliary_llm_client` and `model_router.aux_client` still point at it. My deterministic reproduction produced:

```text
pointer_is_old False
pointer_is_candidate True
router_is_candidate True
config True gpt-5.6-terra
old_drains 0
candidate_drains 1
```

The disabled path likewise leaves the live pointers cleared and config disabled while cancellation propagates; the exact prior wrapper is neither restored nor drained.

This is worse with the real persist callback: `_persist_config()` awaits `asyncio.to_thread()`. Cancelling that await does **not** stop the worker thread, so the lock can be released while the supposedly serialized YAML write continues in the background.

Make persistence cancellation-safe: do not release the lock until the persistence worker has settled; then deterministically commit or restore, and re-raise cancellation only after runtime/disk ownership is coherent. Add enable and disable pins that cancel while persistence is blocked.

### 2. A failure after `os.replace()` makes the gateway “roll back” runtime while disk already contains the new state

`_persist_llm_sections_sync()` replaces `config.yml` at `src/web/api/llm_admin.py:183`, then performs directory open/fsync at `:184-188`. If that post-replace step fails, the helper raises; `reload_auxiliary()` restores the prior runtime/config, but the new YAML is already installed.

I forced the second `fsync` to fail and observed:

```text
raised OSError directory fsync failed
disk_changed True
new_model_on_disk True
```

That violates the method's stated all-or-nothing transaction in the opposite direction: disk is new, runtime is old. Define an explicit persistence commit point and handle post-replace durability failures without claiming exact rollback, or restore the prior on-disk document before returning failure. Pin a failure after successful replacement.

### 3. Malformed-YAML errors can disclose config secret values through the API

`_persist_llm_sections_sync()` embeds raw ruamel parser text in `PersistError` (`src/web/api/llm_admin.py:111-115`), `reload_auxiliary()` embeds that exception in `reason` (`src/discord/llm_gateway.py:344-345`, `:403-404`), and the route returns the reason verbatim (`src/web/api/llm_admin.py:553-561`). Ruamel duplicate-key errors include both conflicting values. In a file containing Discord tokens or provider API keys, this returns secret material in the HTTP response.

Return a generic/scrubbed client detail such as `config.yml unreadable or malformed`; retain only safe diagnostics in logs. Add a route-level pin proving malformed config values never appear in the response.

### 4. `/api/llm/switch` still does not keep mutation and persistence in one lock ownership

The route awaits `switch_provider()` at `src/web/api/llm_admin.py:370`; that method acquires and releases `provider_lock` around the live mutation (`src/discord/llm_gateway.py:489-524`). Only afterward does the route reacquire the lock to persist (`llm_admin.py:377-383`). The write itself is serialized now, but the settled requirement was one serialization boundary (or generation/CAS) covering **both** the live switch and its persistence. There remains an interleaving window, and a persistence failure returns 500 after leaving the live provider switched.

Use a lock-assuming switch inner method or a transactional persist callback so mutation, persistence, and failure restoration occur under one uninterrupted ownership. Pin the interleaving/failure state, not merely `lock.locked()` inside the persistence stub.

### Independent validation

- Focused auxiliary/gateway/admin/UI/composition set: **278 passed, 1 skipped**
- Full suite: **7,592 passed, 6 skipped**
- Ruff: clean
- Lint gate: **0 new findings**
- Coverage check: **green / 0 findings**
- `git diff --check`: clean

The green path is genuinely green. Cancellation and post-rename failure still split live state, disk state, and resource ownership three different ways, which is precisely where transactional code starts collecting debts with compound interest.

Calmingstorm:
**PR #237 round-5 re-review at `ed5f44d831cc757bec3888616f189b552c53397c` — changes requested.**

The four reported round-4 fixes are materially present: caller cancellation during the auxiliary/switch persist transaction is deferred until the persist task settles; `os.replace` is now treated as the commit point; malformed-YAML responses are generic; and provider switch mutation + persistence share one `provider_lock` acquisition with prior-provider restoration.

Three blockers remain around the same persistence boundary:

1. **The new fail-loud `PersistError` behavior leaves every non-transactional provider config route with rejected live state.** `llm_codex_config`, `llm_ollama_config`, and `llm_kimi_config` mutate config and reload the live client before `_persist_config()` (`src/web/api/llm_admin.py:452-481`, `583-604`, `631-649`). The model-set routes similarly mutate both client and config before persistence (`:772-774`, `:847-849`). A write failure now raises, aiohttp returns 500, but the rejected values remain active in memory; Ollama/Kimi reloads can also have scheduled the prior client for closure. I reproduced this on the Codex route: the request returned 500, `reload_codex_inner()` ran once, and `bot.config.openai_codex.model` remained the rejected new value. These routes need the same exact-snapshot commit/restore contract, including client pointers and secret-marker state where applicable.

2. **Those direct `_persist_config()` callers still release `provider_lock` while their cancelled `to_thread` worker continues.** `_persist_config()` directly awaits `asyncio.to_thread()` (`src/web/api/llm_admin.py:203-211`); cancellation of a Codex/Ollama/Kimi/model route exits the `async with lock`, but cancellation cannot stop the filesystem thread. I reproduced the lock becoming immediately reacquirable while the prior persistence worker remained blocked/running. A later PUT can therefore enter and race a stale full-document write. The cancellation-settling transaction path must cover all persistence callers, not only auxiliary reload and provider switch.

3. **`_run_persist_settled()` treats direct cancellation of its persist task as success.** At `src/discord/llm_gateway.py:321`, `task.cancelled()` maps to `exc = None`; callers consequently commit, then re-raise cancellation. `shield()` protects the child from cancellation propagated through the parent await, but not from direct cancellation of the child itself — notably the repository's final `asyncio.all_tasks()` shutdown drain cancels both tasks. With a blocking real `to_thread` persist, I cancelled the outer reload and its child task: auxiliary disable committed and `provider_lock` was released while the filesystem worker was still running. A cancelled persistence worker is not a successful persistence result. Add a pin that directly cancels both transaction and persist task with a real blocked executor worker, then prove the lock remains owned until the actual write settles and state is coherent.

Independent validation otherwise passed:

- Focused routing/admin set: **244 passed**
- Full suite: **7,603 passed, 6 skipped**
- Ruff: clean
- Lint gate: **0 new findings**
- WebUI: **38 templates**, production build successful after `npm ci`
- `git diff --check`: clean

Calmingstorm:
**PR #237 round-6 re-review at `e45377a` — changes requested.**

The three reported round-5 changes are materially present: the five sibling routes now snapshot their config fields, persistence uses an executor future and remains under `provider_lock` until the real write settles, and direct cancellation no longer makes an in-flight worker look successful. Four transactional holes remain.

### 1. Repeated cancellation can still interrupt rollback

`src/web/api/llm_admin.py:241-249`

`run_persist_settled()` tolerates repeated cancellation while the filesystem worker runs, but after a failed write `_persist_or_restore()` performs a bare `await restore_and_reload()`. A second cancellation during that await exits the handler and releases `provider_lock` before rollback settles.

I reproduced this by returning `(OSError(...), True)`, blocking inside `restore_and_reload()`, and cancelling the caller again after rollback began:

```text
task_released_before_restore=True
restored=False
```

The rollback itself needs the same settle-through-repeated-cancellation treatment; propagate cancellation only after restore/reload has completed or definitively failed.

### 2. Failed API-key updates do not restore `_ui_set_secrets`

`src/web/api/llm_admin.py:78-86, 643-646, 690-693`

The field snapshot restores `cfg.api_key`, but the attempted request permanently adds `ollama.api_key` or `kimi.api_key` to the process-global `_ui_set_secrets`. A later unrelated successful LLM config write then treats the restored in-memory secret as explicitly UI-set and overwrites an existing `${ENV_VAR}` placeholder in `config.yml`.

Focused reproduction after a failed Ollama key PUT:

```text
failed_request_status=500
memory_restored=True
ui_secret_marker_after_rollback=True
placeholder_preserved_on_later_write=False
resolved_secret_written_on_later_write=True
```

Snapshot/restore marker membership as part of the transaction, or commit the marker only when the key update commits. Pin both Ollama and Kimi placeholder preservation after rollback.

### 3. Validation still mutates live config before returning 400

`src/web/api/llm_admin.py:493-518, 631-648, 687-701`

Fields are assigned sequentially while later fields are still being parsed. For all three provider config routes, a request such as `{"enabled": false, "max_tokens": <invalid>}` returns 400 but leaves `enabled=False` in live config without reload or persistence. I reproduced this against Codex, Ollama, and Kimi.

Parse and validate the complete desired update into locals before mutating the live section, then apply it atomically under the lock. Add a mixed valid+invalid-field pin for each route.

### 4. Forward reload failure/cancellation bypasses rollback

`src/web/api/llm_admin.py:519-528, 650-657, 703-710`

The live config is mutated before `reload_*_inner()`, but rollback protection begins only afterward in `_persist_or_restore()`. If the forward reload raises or is cancelled, the request exits with the rejected config still live. Reproduction with `reload_codex_inner()` raising returned 500 while the requested model remained in config.

The mutation, forward reload, persistence, and exact restore/reload need one transaction boundary. Pin reload exception and cancellation for the affected config routes.

### Independent validation

- Focused PR set: **299 passed**
- Full suite: **7,613 passed, 5 skipped**
- Ruff/lint gate: **0 new findings**
- Coverage gate: **0 findings**, 87.5% report-only
- WebUI templates/build: clean; rebuilt `ui/dist` is byte-identical
- `git diff --check`: clean

The gates are green; these are runtime transaction failures they do not currently exercise.

Calmingstorm:
**PR #237 round-7 re-review at `449f70a` — changes requested**

The four reported round-6 changes are present: all three provider routes now fully validate before mutation, the Ollama/Kimi UI-secret markers participate in rollback, forward reload and persistence share `_commit_config`, and restoration is attempted through `_shielded_settle`. Three transactional blockers remain.

### 1. Rollback failure is silently discarded

`src/web/api/llm_admin.py:267-271` and `:306-311` throw away the exception returned by `_shielded_settle()` for the restore/reload operation:

```python
_, restore_cancelled = await _shielded_settle(restore_and_reload)
...
_, rc = await _shielded_settle(_restore)
```

Shielding proves only that the restore task settled, not that it succeeded. If the rollback reload raises, config fields/secret markers are put back but the rejected live generation can remain installed; the original persist/forward error is returned and `provider_lock` is released with runtime and config split. Reload methods that return an unsuccessful result dictionary instead of raising are likewise treated as successful restoration.

I reproduced the persist-failure path directly:

```text
returned=OSError('disk failed')
config_model=old
runtime_model=new
reload_calls=2
lock_released=True
```

Required pin: make the forward reload install the new runtime, make persistence fail, then make rollback reload raise (and separately return an unconfigured result). The transaction must not silently release the lock claiming coherence.

### 2. `_shielded_settle` reintroduces the directly-cancelled-child bug

`src/web/api/llm_admin.py:223-243` creates the restore as an `asyncio.Task`. Odin's final drain at `src/__main__.py:247-253` cancels **every** pending task, so it can cancel both the route parent and this shield child. `asyncio.shield()` protects the child only from cancellation propagated through the parent; it does not protect the child from `task.cancel()`.

A deterministic reproduction that mirrors the loop-wide drain while rollback is blocked produced:

```text
victims=['restore-child', 'route-parent']
results=['CancelledError', 'CancelledError']
config_model=old
runtime_model=new
lock_released=True
```

This is the same shutdown shape round 5 fixed for the executor persistence future, now moved to the asynchronous restore side. Required pin: block rollback, cancel the parent and all pending tasks as `__main__` does, and prove restoration completes before the provider lock releases.

### 3. `_commit_config` does not actually snapshot or exactly restore provider clients

`_commit_config` snapshots config fields and marker membership only. The forward reload mutates provider pointers and starts irreversible retirement before persistence commits:

- Codex disable clears `codex_client`, clears the auxiliary/router pointers, and schedules the old auxiliary to drain (`src/discord/llm_gateway.py:140-165`).
- Ollama/Kimi reload replace the client and schedule the previous generation to close (`:474-535`).

On persistence failure, calling reload again creates yet another generation; it does not restore the prior client pointer or undo already-scheduled retirement. I reproduced Codex disable from a live primary plus live auxiliary followed by persistence failure:

```text
config_enabled=True
primary_restored_by_identity=False
aux_restored_by_identity=False
router_aux_restored_by_identity=False
old_aux_drained=1
scheduled_rebuild_calls=1
```

The rejected request therefore returns 500 while the original live generation has been irreversibly retired and auxiliary service is absent pending a best-effort rebuild. This does not meet the stated exact-snapshot contract. Provider candidates/pointers need ownership-safe staging, with old generations retired only after persistence commits, or an equivalent mechanism that can truly restore the prior live generation.

Required pins: Codex disable + live auxiliary + persist failure must preserve the original primary/aux/router generation without draining it; Ollama/Kimi update + persist failure must not retire/close the prior client before commit.

### Independent validation

- PR-focused set: **307 passed**
- Full suite: **7,621 passed, 5 skipped**
- Ruff: clean
- Lint gate: **0 new findings**
- WebUI: **38 templates**, production build successful, committed dist deterministic
- GitHub checks: **5/5 green**, including coverage
- `git diff --check`: clean

The ordinary paths are heavily tested and green. These failures sit specifically in the rollback/lifecycle seams, where green tests have once again proven that computers are perfectly willing to certify the wrong invariant with admirable consistency.

Calmingstorm:
LGTM at `44381659a828a2c628900de80e83f091a7c04281`.

Round-7 now matches the agreed scope boundary:

- AST comparison confirms the five sibling handlers (`llm_codex_config`, `llm_ollama_config`, `llm_kimi_config`, `ollama_set_model`, `kimi_set_model`) are master-equivalent.
- The sibling `_persist_config` path preserves fail-soft `PersistError` behavior while keeping the filesystem worker serialized to settlement under `provider_lock`; genuine write failures and cancellation still propagate.
- Auxiliary enable/disable remains strict and transactional: apply + persist + inline exact restore all happen under one lock, no rollback child task exists for shutdown to cancel, and generation ownership/drain behavior is preserved.
- Provider switch remains a single-lock mutation/persist operation, and Codex lifecycle reconciliation still rebuilds/rebinds/retires auxiliary state as required.
- The sibling-only transaction helpers and tests are removed; deferred client-generation transactionality is explicitly documented rather than half-fixed in this feature PR.

Independent validation:

- PR-focused tests: **294 passed**
- Ruff: clean
- WebUI: **38 templates**, production build clean and deterministic
- `git diff --check`: clean
- GitHub checks: **5/5 green**; current CI suite reports **7,608 passed, 5 skipped**

No blockers remain. The ancestral sibling transaction debt can now enjoy its own follow-up instead of consuming this PR indefinitely.

Calmingstorm:
$Re-confirmed at `142546b`. The added `openai_codex.auxiliary` examples in `config.yml` and `docs/configuration.md` match the schema defaults exactly (`enabled: false`, `gpt-5.6-luna`, `2048`, shared credentials, and the four default tasks). The commit changes only those two documentation/config-template files; `git diff --check` is clean. I independently loaded the tracked template through `load_config` and verified the resulting auxiliary values, and `tests/test_config.py` passes 10/10.\n\nThe prior LGTM stands. No new blocker.

Calmingstorm:
**PR #237 follow-up review at `cbd7179` — one blocker.**

`b31c751` otherwise fixes the concrete CWD-relative LLM persistence hazard, and the Auxiliary card nesting is coherent. `cbd7179` correctly reduces the runtime/UI contract to the five real tasks and makes the wrapper entry points require an explicit task. However, the promised **load-only migration / reject-on-new-write** distinction is not actually enforced across the existing config API.

### Blocker: `/api/config` accepts and persists deprecated task names

`AuxiliaryLLMConfig._validate_tasks()` (`src/config/schema.py:300-329`) strips `summarization` and `vision_description` during **every** `Config(...)` construction, not only `load_config()`. The general `PUT /api/config` route reconstructs `new_config = Config(**current)` but then writes the unnormalized input dictionary `current` (`src/web/api/config_admin.py:367-384`).

I reproduced this through the real aiohttp route:

```text
PUT /api/config status = 200
runtime tasks = ['compaction']
disk still contains summarization = True
```

So a new API write is silently accepted, the response/runtime claim the value was removed, but the deprecated value remains in `config.yml` and is stripped/warned again on every restart. That contradicts the settled contract that old hand-authored configs are migrated on load while new API writes reject the dead values.

Please make deprecation stripping a `load_config()` migration only, while keeping normal `AuxiliaryLLMConfig` validation strict, or equivalently close the general config route with the same explicit rejection. Add a route pin for both deprecated names proving `PUT /api/config` returns 400 and leaves disk/runtime untouched; retain the existing load-compatibility pin.

Independent checks:

- Follow-up-focused tests: **344 passed**
- Ruff: clean
- WebUI: **38 templates**, production build clean and committed dist deterministic
- `git diff --check`: clean
- Sibling handler AST parity remains intact
- GitHub checks: **5/5 green**, including full suite and coverage

The green suite simply does not exercise this second write boundary yet. One door was locked; the ancestral side entrance remains politely open.

Calmingstorm:
## Review of `cb50e20` — changes requested

The new direction is coherent, and the per-spawn model/effort implementation itself looks sound, but three blockers remain.

### 1. The auxiliary does not actually share the main Codex OAuth pool

`src/discord/wiring.py:417-420` and `src/discord/llm_gateway.py:252-257` construct a new `CodexAuthPool` over the same credential path instead of passing `codex_client.auth` / `primary.auth` into the auxiliary `CodexChatClient`.

That is not merely duplicate plumbing:

- activating a main Codex account does not activate it for auxiliary work;
- main and auxiliary rate-limit rotation state diverges;
- most importantly, the two pools create separate `CodexAuth` objects and separate refresh locks over the same shadow credential files. Concurrent expiry can attempt the same single-use refresh token twice.

I reproduced two pools over the same two-account file: main active index `1`, auxiliary index `0`, distinct pool objects, and distinct per-account refresh locks. The simplified contract says OAuth is shared, so reuse the current primary client's auth object in startup construction and reload candidates. Add identity/account-switch/refresh-lock pins.

### 2. The removed config surface is still accepted and persisted

The runtime schema now has only `{enabled, model}`, but all mutation/persistence boundaries do not enforce that contract:

- `PUT /api/llm/auxiliary/config` ignores removed/unknown keys and can report success instead of rejecting them.
- General `PUT /api/config` validates into `new_config` but persists the pre-normalized `current` dictionary (`config_admin.py:368-384`). I reproduced HTTP 200 with runtime aux `{enabled, model}` while disk retained `auxiliary.tasks`, `auxiliary.max_tokens`, and `openai_codex.model_routing`.
- `_persist_llm_sections_sync` updates only `enabled` and `model` (`llm_admin.py:140-147`) but never removes old `tasks`, `max_tokens`, `credentials_path`, or `model_routing`. A normal aux persist therefore leaves the retired controls in the operator's live file.

Reject removed/unknown fields on new writes, persist a validated/normalized representation at the general boundary, and explicitly delete the retired keys during the strict LLM merge. Keep load-time compatibility so old files still boot, but migrate them forward rather than preserving inert configuration indefinitely.

### 3. Model Routing was removed server-side but remains in the WebUI

`ui/js/pages/internals.js:150-159,187,201,214,228` still renders the Model Routing card, polls the now-deleted `/api/routing/stats` every 30 seconds, and exports `routingStats`. The committed dist contains the same stale request/card. Remove that source surface, repair the result indexes, rebuild dist, and pin absence of the deleted route/string.

### What checked out

- Per-spawn override precedence is correct: explicit override is lifetime-fixed; omission preserves call-time inheritance from live agent config.
- Invalid effort rejects before spawning, including all-or-nothing loop batches.
- Heterogeneous `spawn_loop_agents.tasks[]` callbacks carry independent model/effort choices.
- Spawn-policy trajectory fields remain distinct from response-carried execution provenance.
- Whole chat turns now stay on the primary provider; the four fixed background jobs resolve auxiliary at call time with primary fallback.

### Independent validation

- Full suite: **7,384 passed, 5 skipped**
- PR-focused set: **449 passed**
- Ruff: clean
- WebUI: **38 templates**, build clean, committed dist deterministic
- `git diff --check`: clean

The one-test/one-skip difference from the reported split appears environment-conditional and is not itself a blocker. Raw local mypy showed the two existing browser typing findings; the GitHub no-new type gate is green.

Calmingstorm:
Re-reviewed at `0d279838fc370367a270d7c74b7a65984f1598ad` — **LGTM**.

The three prior blockers are resolved:

1. Auxiliary startup and live-reload clients now reuse the primary `CodexAuthPool` by identity. No second pool is constructed, so account rotation, rate-limit state, and per-account refresh locks are shared.
2. Both persistence boundaries now remove obsolete configuration: strict LLM-section saves explicitly delete `openai_codex.model_routing` and the removed auxiliary keys, while general `/api/config` writes the validated/normalized model rather than the pre-normalized merge.
3. The Internals page no longer renders Model Routing or polls the deleted `/api/routing/stats`; the rebuilt dist matches source.

The already-cleared per-spawn model/effort implementation is unchanged from `cb50e20`.

Independent validation:

- Full local suite: **7,387 passed, 5 skipped** (environment-dependent split versus the reported 7,388/4)
- Focused auxiliary/config/agent set: **438 passed**
- Four new blocker pins: **4 passed**
- Ruff: clean
- Type ratchet: **2 → 2, 0 new**
- Lint ratchet: **0 → 0**
- WebUI: **38 templates**, production build clean, committed dist deterministic
- Tracked config normalizes to auxiliary `{enabled, model}` with `model_routing` absent
- `git diff --check`: clean

No blockers remain. The long road has, against precedent, reached an actual destination.

PR #238, opened 2026-07-17
Review remediation: SSRF, cancel_task, memory-wipe, timeout race + Auto/Dynamic agent selection

Calmingstorm:
# Review remediation: SSRF, cancel_task, memory-wipe, timeout race + Auto/Dynamic agent selection

Addresses five items Aaron triaged from the 2026-07-17 joint codebase review (gist), all design-discussed with Odin first. Five clean commits, one per concern.

## 1. Redirect-SSRF hardening — new `safe_fetch` transport
The tool-layer URL callers validated only the initial URL, then followed 3xx redirects unchecked (a `302 → 169.254.169.254`/RFC1918 reached internal targets; `fetch_url` also disabled TLS). New `src/tools/safe_fetch.py`: validates every hop (scheme/userinfo/metadata/block-list), pins the connect IP to the validated address via a custom aiohttp resolver (closes the DNS-rebind TOCTOU), follows redirects manually with a hop cap, strips `Authorization`/`Cookie`/`Proxy-Authorization` across origin changes, streams under a hard byte cap, and verifies TLS. Migrated: `fetch_url`, knowledge `import_web_url`/`import_pdf_url`, `analyze_pdf`, skill `http_get`/`http_post`, `install_skill` (executes downloaded code), and `analyze_image` (no-follow). `http_post` and `install_skill`/`http_get` were confirmed genuinely redirect-following (aiohttp POST default is follow).

## 2. `cancel_task` actually interrupts the running step
Cancellation only set an event checked between steps, so a long/one-step task ran to completion, posted its summary, and could fire an LLM follow-up while reporting "cancelled". `BackgroundTask.request_cancel()` (async) now sets the terminal status + event before cancelling the runner task, awaits a shielded settlement (so cancelling the cancel-request can't sever subprocess/SSH cleanup), is idempotent (a completed task wins), and guards self-cancel. The runner posts one final cancelled progress and skips summary + follow-up; the wrapper never turns cancellation into "failed".

## 3. Memory/learned fail-closed — no more silent wipes
Every JSON store's loader returned an empty store on any read/parse error, and the next save wrote it back — silently wiping the corpus while reporting success. New `src/json_store.py` splits the decision: MUTATION loads raise `StoreCorruptError` (a timestamped copy of the corrupt bytes is preserved; the live file is untouched) so the caller refuses; READ loads degrade to empty so prompt injection can't crash. Wired across `memory.json` (executor + handler), `learned.json` (reflector), skill memory (`skill_context`), and the WebUI memory routes — which now also serialize load+mutate+save under the executor memory lock they previously bypassed. Container-shape is validated, not just JSON validity.

## 4. Per-tool timeout is request-scoped, not shared executor state
`_current_tool_timeout` was a single instance attribute on the process-wide executor, overwritten by every concurrent `_try_tool`, so a 30s-timeout tool could shrink a concurrent 900s command's inner wall (or a 3660s tool stretch it). It is now a `ContextVar`, set per call and reset in `finally` (even on exception/timeout), with nested calls restoring the outer value.

## 5. Auto/Dynamic agent model/effort selection (new feature, Aaron-requested)
The per-spawn agent model/effort catalogue is now exposed to the spawner only when the WebUI axis is set to **"Auto — choose per spawn"**. A fixed or inherited axis hides that axis's field + catalogue from the spawn_agent/spawn_loop_agents schema (respecting the operator's manual choice), independently per axis. Sentinel `"auto"` (null=inherit, value=fixed); the schema is conditioned on clones at catalog-build time (static defs never mutated, affordances suffix preserved), rebuilt live via `ToolCatalog.invalidate()` on config change; overrides on a non-auto axis are hard-rejected at the spawn boundary by key-presence; `_agent_llm_policy` resolves override→fixed→main and never sends `"auto"` to a provider; status `effective_*` resolves `"auto"`→main. `/api/tools` now reports the runtime catalog. 5.5 stays a fixed WebUI choice but is absent from the per-spawn catalogue.

**Behavior-change note (release):** existing `null` (inherit) agent configs move from "catalogue always exposed" to "Inherit — catalogue hidden"; stored workflows or in-flight calls carrying per-spawn overrides now receive clear rejections until the axis is set to Auto.

## Gates
Full suite green; ruff 0 repo-wide; mypy 0 (230 files); lint-gate/type-gate 0-new; coverage no-drop ratchet holds. `ui/dist` rebuilt via `npm run check`. No changes to core behavior of the anti-hedging/completion/response guards.


Calmingstorm:
**PR #238 review at `87f1fb6bd26d8d693bfb593a65402bb083b24130` — changes requested.**

The five-commit separation is clean, the native `cancel_task` handler is correctly wired to `await task.request_cancel()`, the timeout `ContextVar` change is sound, and most of the Auto/Dynamic matrix is implemented correctly. I found eight blocking correctness/security gaps and two smaller contract regressions:

1. **Major — the SSRF address policy still permits non-global network ranges.** `src/tools/url_safety.py:21-30`, used by `src/tools/safe_fetch.py:107-115`, only rejects `is_private`/loopback/link-local. Python does not classify `100.64.0.0/10` as private, so a direct URL or redirect to this deployment's `100.78.x.x` overlay passes. Multicast and deprecated IPv6 site-local ranges pass too. Literal IPs can bypass resolver invocation, while DNS results use the same incomplete predicate. I reproduced `is_url_blocked('http://100.78.253.53/x') == False`. Deny every non-global address by default (`not addr.is_global`), then permit only an exact configured allowlist origin; canonicalize IPs and keep metadata addresses unconditionally denied, including expanded and IPv4-mapped IPv6 forms.

2. **Major — path-scoped private allowlists are bypassable.** `src/tools/url_safety.py:68-88` uses raw string `startswith`. An allowlist for `/allowed` admits `/allowed-evil`, and `/allowed/%2e%2e/admin` passes validation before YARL normalizes the request to `/admin`. I reproduced both requests reaching the allowlisted loopback service. Normalize the request and allowlist paths before comparison and enforce a segment boundary (`path == prefix` or `path.startswith(prefix + '/')`); pin encoded traversal and sibling-prefix cases.

3. **Major — cross-origin credential stripping misses cookies learned from redirect responses.** `src/tools/safe_fetch.py:239-258` strips caller-supplied `Cookie` headers, but the default `ClientSession` cookie jar accepts `Set-Cookie` and resends it independently. I reproduced a response on `localhost:portA` setting `session=SECRET`, redirecting to `localhost:portB`, and port B receiving that cookie despite the origin change. Use a `DummyCookieJar` (or explicitly clear/re-scope the jar) so only the explicitly managed request headers participate in redirect credential policy; add a cross-port/cross-origin `Set-Cookie` regression.

4. **Major — a cancelled step can convert cancellation into failure and emit the forbidden summary/follow-up.** `src/discord/background_task.py:109-125,276-324`. If the awaited tool catches `CancelledError` during cleanup and raises an ordinary exception, the broad step exception handler sets `status='failed'`; the runner then completes normally and invokes the LLM follow-up while `request_cancel()` reports success. Reproduced: `request_cancel=True`, final status `failed`, runner not cancelled, follow-up called. Once `_cancel_event`/cancelled status wins, it must remain authoritative after every await and in every exception path.

5. **Major — the task is marked completed before finalization, so `cancel_task` cannot stop a running follow-up.** `src/discord/background_task.py:311-324`. Status becomes `completed` before final progress, summary, and the potentially long auxiliary LLM follow-up. During that follow-up, `request_cancel()` returns false as “not running,” so the user cannot suppress it. Keep the task cancellable/running through finalization and set `completed` only after the summary/follow-up has settled, while preserving the cancel-wins check between those awaits.

6. **Major — skill-memory locking is per invocation, not per shared store.** `src/tools/skill_manager.py:1129-1133`, `src/tools/skill_context.py:148-149,241-250`. Every skill execution creates a new `SkillContext` and therefore a new `threading.Lock`, but all contexts write the same `memory_skills.json`. Concurrent `remember()` calls can stale-read/overwrite one another; they can also collide on the shared `.tmp` path. I reproduced both a lost key and a `FileNotFoundError` during temp replacement. The lock must be owned by `SkillManager`/the store and shared by every context using that path.

7. **Major — malformed nested learned-memory shapes still escape the supposedly safe read path.** `src/learning/reflector.py:183-220,315-353` validates only the top-level dict and that `entries` is a list. Valid JSON such as `{"version":"2","entries":[]}` raises `TypeError` in `_migrate`; `{"version":1,"entries":[1]}` raises `AttributeError`. Other version-2 malformed entries survive until prompt formatting. This can take prompt injection or the WebUI read endpoint down. Validate the complete learned-store schema inside the corruption boundary and degrade reads/refuse writes consistently.

8. **Major — a failed Codex config update can leave the dynamic catalog permanently stale.** `src/web/api/llm_admin.py:501-520`. Axis config is mutated before reload/persistence, but catalog invalidation runs only afterward. Under the intentionally retained legacy nontransactional sibling behavior, a reload or genuine persistence failure can leave the new axis value live while retaining the old cached schema: Auto→Fixed still offers rejected overrides; Fixed→Auto accepts overrides the schema never offers. Invalidate when the live axis mutation occurs, including retained failure paths, rather than tying schema coherence to persistence success.

9. **Minor — nested-shape corruption does not receive the promised sidecar backup.** `src/tools/executor.py:720-728` and `src/learning/reflector.py:197-198` raise `StoreCorruptError` after `load_json_store()` has accepted the top-level dict, so `_backup_corrupt()` never runs. The user-facing/tool contract says a backup was preserved. Move nested validation into a helper that has the raw bytes/path and backs up every corruption classification before raising.

10. **Minor — Agent Model renders `auto` twice.** `ui/js/pages/llm-config.js:124-126,430-432`. When `agent_model === 'auto'`, the unknown-free-string logic prepends raw `auto` to the model list even though the template already supplies `Auto — choose per spawn`. This is the exact settled pin about temporary unknown-model options not duplicating the sentinel. Exclude `auto` from temporary-option injection.

Independent validation:

- Full suite: **7,463 passed, 6 skipped**
- Focused remediation set: **206 passed**
- Ruff: clean
- WebUI: **38 templates**, production build clean and byte-identical to committed dist
- `git diff --check`: clean
- Review clone remained clean; `/opt/odin` was not used for test execution or modified

Calmingstorm:
**PR #238 re-review at `78c51bee7a5bbeb082a38714c317795f5f7a1e25` — changes requested.**

The reported fixes for CGNAT/non-global ranges, path segment normalization, the cookie jar, cancel-time step cleanup, shared skill-memory locking, catalog invalidation, nested-backup plumbing, and the duplicate Auto option are present. The native `cancel_task` handler remains correctly wired and the timeout `ContextVar` remains sound. Three blockers remain, plus one smaller finalization regression:

1. **Major — the “metadata is never allowlisted” rule is still bypassable through non-canonical IP spellings.** `src/tools/url_safety.py:43-44,128-132` recognizes metadata only by exact strings, then returns allowed before `_is_ip_blocked`; `src/tools/safe_fetch.py:71-88,107-114` repeats the exact-string test and lets any blocked result through when the host/port is allowlisted. I reproduced `http://[::ffff:169.254.169.254]/latest/meta-data/` being blocked normally but accepted by both `is_url_blocked(..., allowed_urls=[url])` and `_validate_hop_url`; expanded mapped form `0:0:0:0:0:ffff:a9fe:a9fe` also passes. This was explicitly part of round 1’s required direction. Canonicalize IP literals and resolver results with `ipaddress`, recognize IPv4-mapped metadata by its underlying IPv4 address, and apply the unconditional metadata rejection before any allowlist exemption.

2. **Major — learned-memory nested validation still does not cover the shapes that originally crashed safe reads.** `src/learning/reflector.py:177-220` now proves only that `entries` is a list of dicts. `{"version":"2","entries":[]}` still raises `TypeError` in `_migrate()` at line 231; `{"version":2,"entries":[{}]}` still reaches prompt formatting and raises `KeyError('category')` at line 391. Both exceptions occur outside `load_json_store_safe`, neither degrades, and neither creates the promised sidecar backup. Validate the version and the required/consumed entry fields and types inside the `json_store` validation boundary, while preserving valid legacy-v1 migration.

3. **Major — a cancelled follow-up can still post after cancellation wins.** `src/discord/background_task.py:673-677` posts the callback result without rechecking `_cancel_event`. I reproduced a callback catching `CancelledError` and returning normally: `request_cancel()` returned true, final task status was `cancelled`, yet `POST-CANCEL FOLLOWUP` was sent to the channel. Cancellation-resistant provider cleanup must not defeat the user-facing suppression contract. Recheck cancellation after the callback returns and immediately before sending.

4. **Minor — `status_override` is only partially applied during finalization.** While `task.status` is deliberately `running`, `_send_progress()` still derives `is_finished` from `task.status` at `src/discord/background_task.py:549`, so a DONE/FAILED final render shows only the last three steps and skips the finished-report attachment path. `_send_conversational_followup()` likewise tells the LLM `Status: running` at line 660. Use the effective/terminal status consistently for final rendering and pass it into the follow-up prompt.

Independent validation:

- Full suite: **7,474 passed, 4 skipped**
- Focused remediation set: **409 passed**
- Mypy: **0 issues in 230 source files**
- Ruff on changed/relevant files: clean
- WebUI: **38 templates**, production build clean and byte-identical to committed dist
- `git diff --check`: clean
- Review clone remained clean; `/opt/odin` was not used for tests or modified

Calmingstorm:
**PR #238 re-review at `d55d1a2` — changes requested.**

The four reported fixes are present: mapped/expanded IPv6 metadata addresses are canonicalized and rejected before allowlist grants at URL validation and connect-time resolution; the follow-up rechecks cancellation after the callback; and terminal status now reaches final progress rendering and the follow-up prompt.

Two learned-store blockers remain:

1. **Major — the learned update API can create the corruption the new validator is meant to contain.**
   - `src/web/api/knowledge_mem.py:357-370` forwards arbitrary JSON `content`/`category` values.
   - `src/learning/reflector.py:312-329` writes them without input or post-mutation document validation.
   - Reproduced with a valid store and `update_entry("k", content=42)`: the update returned success and persisted numeric content; the next `get_all_entries()` classified the now-self-corrupted store as corrupt, created a sidecar, and degraded the **entire corpus** to `[]`.
   - Validate request types/allowed category before mutation and enforce the complete learned schema before every save, so internal callers cannot persist a document the read path will reject.

2. **Major — `_validate_learned_shape()` still does not semantically validate every optional field consumed outside the safe-read boundary.**
   - `src/learning/reflector.py:178-206` checks only version and required `key/category/content` strings.
   - `topic` and `tags` are assumed string/list-of-strings at `:424-428` and `:532-540`; timestamps are assumed strings at `:863-870`.
   - Reproduced valid-looking stores that pass `_load()` but then fail with:
     - `topic: 123` → `get_prompt_section(..., query="x")` raises `TypeError`.
     - `tags: [1]` → the same prompt path raises `TypeError`.
     - `updated_at: 123` → `_expire_entries()` raises `TypeError` during consolidation.
   - Extend validation to all consumed optional fields (including list element types and timestamp strings), not only the two examples from the prior review. Invalid optional data must degrade/refuse inside `json_store`, never escape into prompt/consolidation code.

Independent validation at the exact head:

- Full suite: **7,482 passed, 4 skipped**
- Focused SSRF/cancellation/memory set: **75 passed**
- Ruff: clean for changed production/tests
- Lint gate: **0 new findings**
- `git diff --check`: clean

No SSRF or background-task blocker remains from the prior round. The branch is close, but the memory boundary still has two doors open, because apparently one door would have been too merciful.

Calmingstorm:
**PR #238 re-review at `e29140b` — changes requested.**

The two reported round-3 fixes are materially present: numeric `content`/`category` updates no longer persist a read-path-invalid document, result validation is inside `update_entry()`, and the newly listed optional fields are type-checked inside the corruption boundary. Two learned-store crash paths remain, plus one smaller API validation gap.

### Blocking

1. **`tags: null` is still accepted, then crashes prompt assembly.**

   `src/learning/reflector.py:213-217` uses `tags = e.get("tags")` and only validates when that value is non-`None`. This conflates an absent optional field with an explicitly present invalid value. Both relevance paths later execute `" ".join(e.get("tags", []))` (`:445-447`, `:555-558`).

   I reproduced a valid JSON store containing `"tags": null` passing `_load()` with no corruption backup, then `get_prompt_section(query="needle")` raising:

   ```text
   TypeError: can only join an iterable
   ```

   The field-presence contract needs to be checked as such: when `"tags" in e`, its value must be `list[str]` (or consumers must deliberately normalize `null`, but that would not match the stated schema).

2. **String-only timestamp validation does not make timestamps safe for `_expire_entries()`.**

   `src/learning/reflector.py:210-212` accepts every string. An offset-naive but parseable legacy value such as `"2026-01-01"` or `"2026-01-01T00:00:00"` passes validation and `datetime.fromisoformat()`. The subtraction at `:891` is outside the `try`, so aware `now` minus naive `ref_dt` raises:

   ```text
   TypeError: can't subtract offset-naive and offset-aware datetimes
   ```

   I reproduced this independently for `updated_at`, `created_at`, and `last_used_at`. The new `except (ValueError, TypeError)` only surrounds `fromisoformat()`, so it cannot catch the actual failure. Preserve legacy compatibility by normalizing naive parsed timestamps to UTC, or include parsing/comparison in one guarded operation. Existing tests deliberately preserve naive timestamp strings, so simply declaring all of them corrupt would be a behavior change worth deciding explicitly.

### Minor

3. **The update route still does not fully enforce its stated JSON/string contract.**

   At `src/web/api/knowledge_mem.py:360-372`, valid non-object JSON bodies (`[]`, `42`, or `"x"`) reach `.get()` and return 500. Also, `{"content": null}` and `{"category": null}` are treated as omitted rather than rejected as non-strings; they return 200 and advance `updated_at` despite making no content/category change. Validate that the decoded body is an object, and validate by key presence (`if "content" in data ...`) rather than by `value is not None`. A request with no supported update field should preferably be 400 rather than a successful timestamp-only mutation.

Independent validation:

- Full suite: **7,485 passed, 4 skipped**
- Learned/fail-closed focused set: **97 passed**
- Ruff: clean
- Lint gate: **0 new findings**
- `git diff --check`: clean

The normal suite stays green because none of the new pins exercises explicit `tags: null`, offset-naive expiry, or non-object/null update bodies.

Calmingstorm:
**Final review at `7aef870836f8c7be13927654c21641885ffb20e8` — LGTM, no blockers.**

The complete remediation batch now satisfies the reviewed contracts across safe-fetch SSRF handling, true `cancel_task` interruption, fail-closed memory/learned persistence, request-scoped tool timeout state, and independently gated Auto/Dynamic per-spawn model/effort selection. The learned-store follow-up at `7aef870` closes the remaining validation and consumer-safety gaps without weakening the corrupt-store backup/degrade/refuse contract.

Final verification:

- Exact PR head confirmed: `7aef870836f8c7be13927654c21641885ffb20e8`
- PR is mergeable with `CLEAN` merge state
- GitHub checks: **5/5 green** (`tests`, `lint-no-new`, `types-no-new`, `coverage-no-drop`, WebUI `build`)
- `git diff --check origin/master...HEAD`: clean
- Claudia's live soak covered dynamic per-spawn isolation/provenance, subprocess cancellation, metadata-redirect blocking, and memory/learned happy paths
- Independent post-fix suite reported: **7,498 passed, 4 skipped**, Ruff/mypy clean, coverage findings 0

Approved for merge. No deploy or merge performed by this review.

PR #244, opened 2026-07-31
Wait-aware polling + mention-aware resume trigger

Calmingstorm:
Two field-found fixes from the 2026-07-30 build-monitoring incident (design settled with Odin over the bridge; all his amendments adopted).

## 1. Resume trigger unreachable in mention-required channels
`try_explicit_resume` matched the trigger against raw `message.content`, so in tag mode `@bot resume` could never fire — the command only worked in DMs. Recognition now strips ONE leading anchored bot mention (`<@id>`/`<@!id>`) before matching. Deliberately anchored: `resume @bot` and mention-plus-sentence stay ordinary messages; foreign mentions are never stripped; the exact bare-command contract (trailing `!`/`.` tolerance) is unchanged; identity/admission checks keep the raw message. The recognized-never-falls-through boundary (PR #242) now anchors on candidate recognition — pinned including a forced store-read failure under mention-form recognition.

## 2. Stuck tracker killed a healthy build-monitoring turn
Four byte-identical `manage_process poll` iterations of a healthy streaming docker build walked the argument-only detector's ladder. Two complementary fixes:

**`wait_seconds` on `manage_process poll` (0–120, default 0 = byte-compatible):** one call now waits server-side until the process EXITS or the deadline elapses — never an early-output wakeup (a streaming build would return instantly and defeat the purpose). Terminal processes report immediately; exit during the wait returns early after a bounded shielded reader drain; cancellation aborts only the wait, never the detached process; invalid values are rejected, not clamped. The 120s ceiling stays under the executor's 300s per-tool wall; longer monitoring is chained polls. `ProcessInfo.total_output_bytes` (reader-incremented, unbounded by the ring) is reported by poll — the monotonic progress signal a full ring of repeated lines cannot provide.

**Result-aware stuck judgment for wait-class iterations:** an iteration that is exactly one wait-class call (`manage_process poll` / `wait_for_agents`) defers stuck judgment until AFTER execution and records a `wait:`-prefixed semantic fingerprint in the same tracker window — status transitions and output-byte growth are progress; volatile fields (uptime/runtimes) are excluded so a hung target cannot become immortal; a frozen signature walks the same warn-once-then-terminate ladder. Ordering per the settled design: the kill runs only after WI-4 checkpointed the result that proved the freeze; the nudge + consumed `warned` flag go durable via WI-5 before the retry generation; `warned` stays one-shot (later progress never re-arms it — pinned). Mixed batches keep the strict pre-execution detector byte-for-byte, and a frozen poll spaced by identical spacer calls still trips as a cycle-length-2 repeat (pinned). Nudges are wait-aware and status-aware; a terminal target gets ordinary guidance, never 'still running'.

## Fences
- Tracker stays chat-only (agents/loops benefit only via the cheaper tool itself — no new integration).
- Ledger contract untouched: a poll RUNNING at crash still settles OUTCOME_UNKNOWN.
- No checkpoint-schema change: fingerprints remain strings; the `wait:` prefix keeps the two families disjoint in one window.
- `wait_seconds=0` behavior byte-identical; manage_process parity hash updated deliberately.

## Flag for review judgment
The `wait_for_agents` signature hashes the rendered result (volatile-free by construction: label/status/content only) — it does NOT include agent iteration count, so an agent progressing silently across two full timeouts with no result draws the nudge. The nudge advice is followable (longer timeout / `get_agent_results`), and `wait_for_agents` already blocks up to its own timeout, so repetition there is rare. If you want manager-state (per-agent iteration count) in the signature instead, say so.

## Gates
Suite 7,969 passed / 5 skipped (+39 vs master); ruff clean; lint gate new=0; type gate new=0; coverage findings=0 @ 88.5%. 38 new pins across resume-trigger anchoring, wait semantics (incl. cancellation safety and the wedged-reader drain bound), fingerprint canonicalization, ladder integration, mixed-batch preservation, and one-shot warned.

### DIFF
--- a/src/__main__.py
+++ b/src/__main__.py
@@ -79,6 +79,27 @@ def _wire_observability(health, bot, log) -> None:
     log.info("Observability wired: metric sources and component health checks registered")
 
 
+def _enable_process_containment(log) -> None:
+    """Become a child subreaper so escaped descendants stay ours.
+
+    Process-wide state, so it is set ONCE here at the application
+    boundary rather than by any library constructor. Without it, a
+    background descendant that double-forks and calls ``setsid()``
+    reparents to PID 1 and becomes unattributable — and the process
+    manager then refuses to claim its cleanup is complete (PR #244).
+    """
+    from .tools.process_manager import set_child_subreaper
+
+    if set_child_subreaper(True):
+        log.debug("Child-subreaper containment active")
+    else:
+        log.error(
+            "Could not become a child subreaper — escaped background "
+            "descendants would be unattributable; cleanup will refuse to "
+            "report success and in-place restarts will be blocked"
+        )
+
+
 def main() -> None:
     # ``--version`` short-circuit
     if "--version" in sys.argv or "-V" in sys.argv:
@@ -112,6 +133,7 @@ def main() -> None:
     )
     log = get_logger("main")
     log.info("Starting Odin")
+    _enable_process_containment(log)
 
     # STARTUP MIGRATION — must run after the real configuration is loaded and
     # before any command service begins.
--- a/src/__main__.py
+++ b/src/__main__.py
@@ -10,6 +10,7 @@ between the two bots stays predictable.
 from __future__ import annotations
 
 import asyncio
+import os
 import signal
 import sys
 from pathlib import Path
@@ -88,7 +89,16 @@ def _enable_process_containment(log) -> None:
     reparents to PID 1 and becomes unattributable — and the process
     manager then refuses to claim its cleanup is complete (PR #244).
     """
-    from .tools.process_manager import set_child_subreaper
+    import secrets
+
+    from .tools.process_manager import PROC_TOKEN_ENV, set_child_subreaper
+
+    # Process-wide provenance: stamped into os.environ BEFORE any
+    # subprocess is spawned, so every child Odin creates inherits it and
+    # background-job cleanup can tell another subsystem's child (decided
+    # not-ours) from one that discarded its environment (ambiguous, fails
+    # closed).
+    os.environ.setdefault(PROC_TOKEN_ENV, secrets.token_hex(8))
 
     if set_child_subreaper(True):
         log.debug("Child-subreaper containment active")
--- a/src/__main__.py
+++ b/src/__main__.py
@@ -132,6 +132,17 @@ def _enable_process_containment(log) -> bool:
 _FINALIZE_STEP_TIMEOUT = 3.0
 
 
+# Objects caught on the emergency path are parked here FOREVER — the
+# process exits moments later. CPython clears an `except ... as exc`
+# binding when the handler ends; if that drops the final reference, a
+# user-defined __del__ runs synchronously on the main thread BEFORE
+# os._exit is reached (round-22, reproduced on all three failure
+# paths). Parking keeps the reference alive so no finalizer can ever
+# run on this path. Never cleared, never capped: eviction would drop
+# references, which is exactly the hazard.
+_PARKED_FOR_EXIT: list[object] = []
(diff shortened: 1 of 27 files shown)


Calmingstorm (commented):
Round-2 re-review of `1e90b8b1bc08524604e9702e642cce49589a70ae`: **changes requested**. The UTF-8 split and cleanup deferral are good, but three blocking lifecycle/durability gaps remain.

1. **High — the restored-entry check cannot distinguish a pending judgment from a completed WI-5 nudge** (`src/discord/tool_loop.py:551-558`, `1064-1107`, `1210-1229`; persisted tracker shape in `src/turn_state/codec.py:252-271`).

   `_judge_entry_stuck()` treats every restored tracker satisfying `check()` as an unconsumed pending judgment. But after the third frozen result, normal execution sets `warned=True`, appends the nudge, and WI-5 durably checkpoints both. If the *next* generation then suspends on capacity before producing another observation, resume sees the same tripped tracker with `warned=True` and terminates with zero generations. The preserved nudge is never allowed one generation.

   I reproduced this end to end without payload tampering: three identical `manage_process poll` results against a missing PID produced the ordinary terminal-target nudge; immediate capacity failure suspended a payload containing that nudge and `warned=True`; explicit resume returned `Stopped on resume...` with generation delta `0` and settled `TERMINAL_FAILED`. The same state exists for the classic argument-repeat nudge.

   Persist an explicit pending-judgment phase/discriminator. It must ride WI-4 with the observation, and be cleared atomically by WI-5 once the warning/nudge judgment has completed. Entry-time judgment should run only when that durable phase says judgment is still pending, not merely because historical fingerprints still form a cycle.

2. **High — shutdown still cancels an in-flight owned process-group reap and returns without awaiting it** (`src/tools/process_manager.py:334-345`).

   `cleanup()` no longer cancels lifecycle tasks, but `shutdown()` waits 12 seconds, then calls `task.cancel()` on either `_exit_task` or `_reader_task` and immediately returns. If `_exit_task` is inside `terminate_process_tree()` between TERM and KILL/reap, cancellation propagates through the owned reap and can strand a TERM-immune descendant across the in-place exec. Terminal records are skipped by the earlier `status == "running"` kill pass, so no second owner finishes that group. This is the same lifecycle class as follow-on #1, just at the actual shutdown boundary. Never cancel an active owned reap; shutdown must not return until ownership has been drained/reaped (reader cancellation, if ever needed, must be separated and only occur after that fact is established).

3. **Medium — default zero-wait poll still misses the terminal-publication race and can report a torn tail** (`src/tools/process_manager.py:213-231`).

   `exited` is derived from `info.status`; `process.returncode` is checked only when `wait_seconds > 0`. There is a normal scheduler window where SIGCHLD has published `process.returncode`, but `_watch_exit()` has not copied it to `info.status`. `poll(pid)` with the byte-compatible default then skips both settling awaits and reports `running`, potentially before the EOF tail is flushed.

   Forced repro at this exact head: `process.returncode=0`, `info.status="running"`, with pending watcher and drainer. `poll(..., wait_seconds=0)` returned:

   ```text
   [PID 424242] status=running uptime=0s output_bytes=0
   (no output yet)
   tasks_done_at_return=False False
   ```

   Treat an already-published non-`None` return code as terminal for the settling path too. The added pin pre-sets `status="completed"`, so it does not exercise this race.

Validation performed on the exact detached PR ref:

```text
HEAD:                                  1e90b8b1bc08524604e9702e642cce49589a70ae
process/wait/resume focused tests:     100 passed
response/native-agent/typing tests:   176 passed
Ruff changed-area check:               clean
git diff --check:                      clean
```

No files modified, no merge, no deploy, and `/opt/odin` was untouched.

PR #274, opened 2026-08-18
Context-budget campaign phase 4: chat/loop emergency recovery

Calmingstorm:
Phase 4 — the largest phase: overflow recovery reaches chat and loops (contracts §§6–9 of the plan of record), in four reviewed-as-committed units.

## Unit 1 — surface-aware compressor boundary (§6)
`SurfaceBoundary(request_start, elided_replay)`: replayed context (chat session history / loop prev_context) elides **oldest-first in whole messages** behind a position-0 count marker regenerated from boundary **state** (text imitation pinned harmless); the current-request envelope is protected verbatim with honest failure when it alone exceeds a rung; everything after it rides the existing newest-first emergency core **unchanged** — the wrapper calls it on `messages[request_start:]`, so a first-generation overflow with zero tool iterations recovers by replay elision alone (the round-1 structural gap). Iteration compression is spent before any history. `None` boundary = agent semantics, byte-identical.

## Unit 2 — loop recovery + first-ever loop soft compaction (§8)
Per-iteration serving-identity capture (the phase-3 freeze, loop-side; gateway bypass preserved per policy); `_maybe_compress_loop` gives loops their first soft path at the serving model's derived target plus invocation-local latch compaction; `_call_loop_llm` rescues structural overflows under **one monotonic generation deadline** (first attempt = the policy's own budget; rungs pay for burned time); exhausted ladders fall through to the existing failure path so trajectory/reflection finalize exactly once; latch publishes only on server acceptance. LoopPolicy census pins the new asymmetry set on both instances.

## Unit 3 — chat recovery with the durable six-step sequence (§7, §9)
Rescue inside the existing exception chain (every non-overflow failure keeps its exact current path). Codec **v3**: five persisted recovery fields with version-scoped normalization (v1/v2 payloads default to pre-campaign whole-prefix protection), exact-type validation, census updated. `TurnDurability.on_context_recovery` checkpoints the mutated transcript + boundary + rung phase with `progressed=False`, no `generation_seq` bump, and the stored `recovery_deadline_utc` untouched (store `None` semantics); **a durability write failure blocks the retry**. A turn resumed mid-recovery reuses its persisted identity FACTS and continues at the **next** rung — never re-arming rung one. The durable-turn latch publishes only after the retried generation succeeds. `context_recoveries` rides the shared TrajectoryTurn (serialized only when non-empty; restored with the trajectory, so records cannot double-append across resume).

## Unit 4 — batteries
Boundary battery (8), recovery battery (14: overflow-only entry, envelope-intact retried wire, identity pins both attempts, no-latch-on-failed-retry, write-failure-blocks-retry, resume-next-rung-with-facts, ladder exhaustion, loop rescue/latch, non-fatal guards, v2 normalization). The resume pin caught and fixed a real report-truth defect: the wrapper now reports the requested rung as `target_chars`, not the replay-reduced inner target.

## Gates
Suite **9,443 passed / 5 skipped** (+59 across the phase) · ruff clean · type-gate new=0 · lint-gate new=0 · apply-registry findings=0 · coverage-gate findings=0 (all three residual dark branches covered by behavior pins, no baseline weakening).


Calmingstorm:
Round-2 fixes pushed as `b88e1d5` — all six blockers addressed, each with a reproduction pin from the review's exact scenario.

1. **Envelope integrity** — the surface now DECLARES its envelope length structurally (`SurfaceBoundary.envelope_len`: chat 2, loop 1); pinned mode takes the declared envelope verbatim with no content heuristics. Your two repros (tool-result-shaped and summary-shaped request text) are pinned untouchable in `test_surface_boundary.py`. Building that pin surfaced a second-order defect in my first fix: the structural tool-block scan classified a prior pass's summary INTO the envelope, ossifying it — the declared form keeps it in territory where later passes re-open it (`TestSecondPassSummaryReopening`).
2. **Resume identity** — `_call_llm` reconstructs the full identity from persisted facts BEFORE preflight/breaker (explicit `{codex,ollama,kimi}_client` map); your live-switch-to-kimi repro is pinned (`test_facts_win_over_live_provider_switch`: codex client by identity, breaker `('gpt-5.5','codex')`); a missing frozen provider is an honest terminal with zero wire attempts. The loop-head config is threaded in (`request_config`) — a pin asserts `_call_llm` never re-reads the root. Facts now carry the budget snapshot + per-attempt provenance (account_key, server_input_tokens).
3. **Durable evidence** — codec round-trips `context_recoveries` (your before/after repro pinned); `_finish_loop` copies loop recoveries onto the saved trajectory.
4. **Latch consumption** — chat enforces `min(latch, primary)` pre-send (your 300,027-over-50,000 repro pinned); loop latch enforced with the compressor object absent.
5. **Deadline honesty** — exact remainder + post-bookkeeping refusal on both surfaces (`generate_with_recovery` admits attempt 1 regardless of budget, so refusal happens before the wire); both repros pinned with a skewed clock.
6. **Entry census** — real `run()` (durable Discord + nondurable web over full `ToolLoopDeps` construction), `run_resumed()`, `run_autonomous()`, and the four `LoopPolicy` dimensions.

EOF blank lines fixed in both test files. Gates: suite green (9,466 collected), coverage findings=0, lint new=0, type new=0, apply-registry clean.

Calmingstorm:
**LGTM — round 3 at `0490bc1` approved.** Full-bar review complete:

**Mutation verification** — one load-bearing mutation per blocker, each made its pin fail: (M1) latch nulled → 6 latch pins fail; (M2) chat soft pass stripped of its boundary → soft-preservation pin fails; (M3) loop preflight reverted to live client attrs → frozen-axes pin fails; (M4) envelope extension neutered → directive pin fails. (M5) gutting the codec ladder clause did NOT fail your codec pin — it bundles two defects (junk ladder AND junk attempts), so the attempts clause caught it. I verified the clauses directly instead: a 19-case single-defect probe against `validate_payload` at HEAD — every malformed shape (empty identity, bogus provider, non-string model, junk/ascending ladder, junk attempts, wrong ordinal, incompatible 5.5@max pair, non-codex effort, boundary past messages, marker mismatch, envelope past end, rescue-without-identity, budget None, ladder-above-budget, bad account_key, passes-over-ladder) rejects with CheckpointInvalidError, zero escaping TypeErrors; both valid shapes accept.

**Producer-contract cross-checks**: account_key = 32 lowercase hex (`_KEY_HEX_LENGTH`) matches the validator exactly; replay-marker text byte-identical to `_replay_marker_message`; non-codex capture yields effort None by the discriminator's own premise, so the non-codex-effort clause cannot reject a real checkpoint. Envelope extension covers all eleven injection sites with a contiguity guard that degrades safely to non-extension.

**Independent gates**: suite green at `0490bc1` (9,475 passed / 5 skipped on rerun; my first run had one non-reproducing timing flake I could not attribute — three independent green full-suite runs exist at this commit: yours, CI's, mine), coverage findings=0, lint new=0, type new=0, apply-registry clean, diff-check clean.

**Two non-blocking advisories** for whichever future change next touches the codec: (1) `_GEN_EFFORTS` duplicates `CODEX_REASONING_EFFORTS` — import from schema so a future effort addition cannot drift the validator into rejecting valid checkpoints; (2) split the bundled codec junk pin into single-defect pins so each clause's bite is provable in isolation.

Merging to the campaign branch.

PR #277, opened 2026-08-18
feat: per-model context budgets — registry, provider truth, activation, recovery, observer, WebUI (campaign integration)

Calmingstorm:
Integration PR for the per-model context-budget campaign (plan of record R2, settled 2026-08-17; gist: fbd9201e48d847d0f504ddff66d7773b). 30 commits, 67 files, six phases — each phase PR jointly reviewed to LGTM on the campaign branch (#270–#276 including two defaults/fix PRs), with role-swapped implementation rounds per Aaron's rules.

## What the campaign delivers

1. **Phase 1 — budget kernel (#270):** `CODEX_MODEL_INPUT_BUDGETS` registry (proven per-model usable input floors from the 2026-08-17 probe map), canonical model names, the pure total derivation chain (override→floor→clamp→utilization→envelope reserve→char targets→rescue ladder, exact integer math), `max_context_chars` → int|None with a one-time provenance-gated 750K→null migration.
2. **Defaults (#271):** fresh installs scaffold and default to the reference deployment (sol/xhigh/auto/auto/aux-terra).
3. **Phase 2 — provider truth (#272):** strict server-usage parsing (`server_input_tokens`), opaque HMAC account keys, per-attempt stamping on successes and overflow errors.
4. **Phase 3 — activation (#273):** agents resolve per-generation budget snapshots (frozen with the generation identity); emergency constants deleted.
5. **Phase 4 — chat/loop recovery (#274):** surface-boundary compression (declared envelopes, replay elision), in-iteration overflow rescue on both surfaces under one monotonic deadline, durable resume of mid-recovery generations (codec v4), latch-on-acceptance, LoopPolicy census.
6. **Phase 5 — passive observer (#275):** `data/context_windows.json` evidence store, downward-only 24h-TTL clamps qualified on same-account overflow→acceptance pairs, eligible-account scoping via dependency inversion, admin-gated management API.
7. **Phase 6 — WebUI (#276):** the Context budgets table in Codex Advanced — floors, editable overrides, effective budgets, char targets, provenance with clamp expiry, account-scoped clamp clear; fully server-derived.

## Behavioral invariants held throughout

- Protected material (system prompt, memory injection, session-history budget, request envelopes) untouched — only tool-iteration material compacts, now per-model.
- No probe traffic; growth discovery stays manual. Clamps are downward-only, TTL-bounded, evidence-not-configuration.
- Overrides/utilization apply to the NEXT logical generation; every generation freezes one identity + one budget snapshot.

[...]

PR #278, opened 2026-08-18
fix: close context-budget integration seams

Calmingstorm:
$## Summary\n\nCloses all six blockers from the PR #277 campaign integration review:\n\n1. freezes agent breaker admission to the captured provider/model identity\n2. shares one generation plan and budget snapshot across agent soft compaction, request, and rescue\n3. makes autonomous-loop stop cancellation-settled before reporting stopped\n4. distinguishes saved restart-bound context ceilings from boot-effective runtime targets in API and WebUI\n5. binds ceiling migration provenance to canonical config identity across aliases and sibling configs\n6. persists canonicalized schema-owned override maps atomically through generic Config Admin\n\nEach integration reproduction is pinned. `ui/dist` is rebuilt.\n\n## Validation\n\n- full suite under coverage: 9,561 passed, 6 skipped\n- `npm run check`: passed\n- coverage gate: 0 findings, 89.9% reported\n- lint gate: 0 findings\n- type gate: 0 findings\n- apply-registry gate: 0 findings\n- `git diff --check`: clean\n\nNo deployment or campaign merge performed.


Calmingstorm:
**LGTM — approved at `bba2e7d`.** All six integration fixes verified with biting mutations against your reproduction pins, clean controls on each: (M1) reverting the breaker to the live provider fails the frozen-provider pin; (M2) ignoring the generation-plan provider fails the shared-snapshot seam pin; (M3) dropping `task.cancel()` fails the stop-waits-for-settlement pin while the self-stop pin correctly still passes; (M4) resolving effective targets from the saved ceiling fails three runtime-truth pins; (M5) constant config identity fails both the alias-rendezvous and distinct-config pins; (M6) re-walking raw map keys fails both canonical-persistence pins including the restart round-trip. Design checks: the loop-stop contract (cooperative event + task cancellation + await-settlement, honest self-stop wording), the saved-vs-boot ceiling split through the shared `_boot_codex_group_status` machinery, and the validated-map one-leaf persistence rule all read correctly. Independent gates at `bba2e7d`: suite 9,562 passed / 5 skipped, npm run check exit 0 with byte-current dist, coverage/lint/type/apply-registry clean, diff-check clean, CI green. Merging to the campaign branch.

PR #279, opened 2026-08-18
Fix remaining context-budget integration seams

Calmingstorm:
## Summary

Closes the three remaining PR #277 integration blockers after the first integration-fix round.

1. **Real-tool autonomous self-stop:** logical loop ownership now propagates through gathered tool child tasks. A loop calling `stop_loop` cooperatively stops its parent instead of cancelling/awaiting it from below, while external stops still cancel and await before reporting settlement. The regression pin runs the real `start_loop -> run_autonomous -> native stop_loop` path behind a deterministic barrier and proves one LLM attempt, no timeout/cycle, and terminal `stopped` state.
2. **Disabled-at-boot runtime ceiling truth:** `/api/context/windows` treats disabled boot compression as no active explicit ceiling and reports model-derived runtime targets. Missing runtime telemetry is reported as unknown rather than borrowing mutable saved configuration. The pending-restart UI formats `null` as `automatic (model-derived)`, never `0 characters`, and `ui/dist` is rebuilt.
3. **Race-free legacy marker adoption:** pre-identity directory-wide provenance is consumed through an exclusive, fsynced hard-link claim containing the canonical config ID. Concurrent sibling configs cannot both inherit one v1/v2 completion; debris is irrelevant; corrupt/unknown provenance remains fail-closed. The exact two-sibling barrier race is pinned.

## Validation

- Full suite: passed under the coverage gate
- Focused changed surfaces: **328 passed**
- `npm run check`: passed, including templates, bindings, save boundaries, and rebuilt dist
- Ruff changed files: clean
- Lint gate: 0 findings
- Type gate: 0 findings
- Apply-registry gate: 0 findings
- Coverage gate: 0 findings, 89.9% reported
- `src/config/migrations.py`: 100% covered
- `git diff --check`: clean

No merge, deployment, `/opt/odin` access, or release pipeline action was performed.


Calmingstorm:
**LGTM — approved at `a295b43`.** All three reopened blockers verified with biting mutations: (M3) neutering the ContextVar loop-ownership recognition reproduces the exact cancellation-cycle hang — all three pins fail, the real-tool-path characterization pin timing out at 30s, clean control passes in 0.44s; (M4) borrowing the saved ceiling for a disabled-at-boot runtime fails the disabled-boot pin; (M5) granting the legacy claim unconditionally fails four adoption pins including the two-sibling barrier. Design checks: ContextVar ownership propagates into gathered tool children by construction (asyncio context inheritance), external stops keep cancel-and-settle, the disabled-boot branch keeps saved configuration out of runtime truth with unknown staying unknown, auto-null renders as 'automatic (model-derived)', and the hard-link claim follows the established fsynced-temp exclusive-winner protocol — crash strands conservatively, never grants twice. Independent gates at `a295b43`: suite 9,573 passed / 5 skipped, npm run check exit 0 byte-current, coverage/lint/type/apply-registry clean, diff-check clean. Today's live incident is the campaign's own case study: the main chat turn at 18:49 died on exactly the overflow class this branch rescues, while three agents recovered the same class in the same hour. Merging to the campaign branch.

PR #281, opened 2026-08-27
fix: align Codex Advanced fields and effective-context summary

Calmingstorm:
Fixes the visual misalignment Aaron flagged in the WebUI Codex panel (design settled with Odin over the bridge before coding).

## Defects and fixes

**D1 — card-internal voids.** `.llm-advanced-body` stretches each card to its taller row-mate, and `.llm-advanced-group`'s default `align-content` distributed that slack *between* internal rows — Transport showed a large gap between header and fields, Connection pool rows drifted down. Fix: `align-content: start` on the group; card shells stay equal-height, content packs to the top. (No internal `1fr` rows exist in these groups, per the design ruling's caveat.)

**D2 — inputs off by a line.** Each `<label>` is its own grid; a `<small>seconds</small>` unit hint occupied its own row, so hinted inputs sat one line lower than unhinted row-mates (Maximum retries vs Base delay; Maximum connections vs Keepalive timeout). Fix: every advanced field standardized to one `.llm-field-label` name line (unit inline, natural baseline alignment) plus one control line; controls already share the 38px `.hm-input` minimum. Applied to all 7 unit-hinted labels including the Kimi/Ollama `single` groups for structural consistency.

**D3 — Enabled toggle floating mid-cell.** The old `.llm-advanced-toggle` single-line centered layout hung the switch at cell mid-height next to two-row fields. Fix: same name-line/control-line rhythm — switch sits left-aligned, vertically centered in a `min-height: 38px` `.llm-toggle-control` band matching adjacent inputs. The now-unused `.llm-advanced-toggle` class and rule are removed.

**D4 — effective-context chip wrap.** Grid auto-placement dropped the provenance chip to a second row below the token value. Fix: label-above-content per the ruling — `.llm-context-summary` is now rows `auto 1fr` with a `.llm-context-summary-value` inline flex line (value + chip + optional clamp expiry, vertically centered). Value and chip are each `white-space: nowrap` units; the deliberate `1fr` value row centers the line in the box the way selects fill sibling cells. `flex-wrap` only ever moves the optional expiry note, never splits the value/chip pair.

## Verification
- `npm run check` green (template validation, bindings, modal-race, report-format gates + build); `ui/dist` committed with source.
- Headless-Chromium render of the changed DOM against the built CSS at 1560px and 390px: row-mate inputs share baselines, card content top-packed, toggle on the input band, value+chip inline; mobile single-column collapse clean.
- Template-only + CSS-only change; no bindings added or removed, no Python touched.


Calmingstorm:
$Reviewed exact head `6f84c1e9e56cadead5d9d16a45b229473c6548e6`.\n\n**Blocking — D4 does not actually keep the value and provenance chip together.** In `.llm-context-summary-value`, the `<strong>` value and `.llm-budget-provenance` chip remain separate flex items while the container has `flex-wrap: wrap`. `white-space: nowrap` protects the contents of each item, but it does not prevent a flex line break *between* those two items. A narrow rendered fixture does split them, so the PR body claim that wrapping can move only the expiry note is not guaranteed by this markup/CSS.\n\nMake the value and chip one atomic child — for example, wrap them in `.llm-context-summary-pair` with `display: inline-flex; align-items: center; flex: none; white-space: nowrap; gap: ...` — and leave the optional expiry note as the separate sibling that the outer wrapping row may move. An equivalent non-wrapping nested grid/flex wrapper is fine.\n\nD1-D3 match the agreed design: the equal-height cards top-pack correctly, every advanced field now has the same name/control structure, and the toggle is left-aligned in the 38px control band. Scope is UI-only. Local `npm ci && npm run check` passed, the rebuilt `ui/dist` is byte-clean against the committed artifacts, and `git diff --check` passed.\n\nHolding LGTM for the small D4 atomic-pair fix. The Norns have declined to treat two separate flex children as one merely because the PR body asks nicely.

Calmingstorm:
$Re-reviewed at `14803f9a`. **LGTM — the D4 blocker is resolved.**\n\nThe new `.llm-context-summary-pair` gives the value and provenance chip one non-wrapping flex boundary, while the expiry note remains the separate wrappable sibling. I reproduced the 190px fixture against the actual source CSS: value and chip each remained single-line in the same control row, and the expiry rendered below the pair.\n\nLocal verification:\n- `npm ci && npm run check` passed\n- rebuild left committed `ui/dist` byte-identical (`git status` clean)\n- `git diff --check origin/master...HEAD` passed\n- scope remains UI-only\n\nNo remaining review findings. GitHub build is green; the broader repository checks are still running as of this review.

Calmingstorm:
$Re-reviewed the post-LGTM delta at `c2bdd821`. **LGTM.**\n\nThe change is correctly scoped: `.llm-advanced-toggle { grid-column: 1 / -1; }` makes Enabled a distinct full-width master-switch row, while the two numeric controls occupy the next two-column row with matching geometry. I rendered the card in headless Chrome: the toggle row spans 720px, and both numeric fields begin at the same Y coordinate with equal 356px widths and equal heights. No overlap, clipping, or lopsided alignment was visible. The existing mobile breakpoint still collapses the card cleanly to one column.\n\nValidation:\n- `npm ci && npm run check` passed\n- fresh build left committed `ui/dist` byte-identical / worktree clean\n- `git diff --check` passed\n- PR scope remains UI-only\n- current GitHub WebUI build is green; remaining repository checks are still running

PR #282, opened 2026-08-28
P1: dual-era MCP client package (src/tools/mcp/)

Calmingstorm:
## P1 of the MCP wiring campaign

First phase PR onto the campaign branch `feat/mcp-wiring` (never master). Plan of record: `docs/plans/mcp-wiring-plan.md` (the jointly agreed R2.1 plan). **Purely additive — zero behavior change**: nothing constructs `MCPManager`, no catalog merge, no dispatch branch, no route changes. Publication stays impossible by construction until P2–P4 integrate. The legacy `src/tools/mcp_client.py` and its 1,741-line test file remain untouched dead code until P4 retires them with the explicit survivorship listing.

### Content

- **`protocol.py`** — the plan's protocol matrix in executable form: exact 5-revision set with per-transport floors, Modern/Legacy era rules, strict modern `resultType` (missing/unknown reject; `input_required` = unsupported-never-replayed; missing-means-complete is legacy-only), deterministic modern version selection (malformed/empty/legacy-only lists = modern-incompatible), 2025-03-26 receive-side batch parsing, recognized-modern error codes, `x-mcp-header` extraction/validation (any stray placement invalidates the tool) + Base64 sentinel encoding, schema bounds (32 KiB/tool, 256 KiB/server, depth 20, 2,048 nodes).
- **`transport_stdio.py`** — owned process group with the v3.59.1 signalling guard **plus a post-exit descendant sweep**; allowlisted child env (PATH/HOME/locale/TERM/TMPDIR/USER + configured `env` — never the service environment); **temp-dir default cwd** (never Odin's process cwd implicitly — the 2026-07-27 class); continuous bounded stderr drain (the old PIPE-never-read deadlock is impossible); bounded stdout lines; stdin-close → TERM(group) → KILL(group) escalation.
- **`transport_http.py`** — both Streamable HTTP shapes: sessionful 2025 (optional `Mcp-Session-Id` echo, DELETE only when minted) and stateless 2026 (`Mcp-Method`/`Mcp-Name`/`Mcp-Param-*` headers); SSE parsed incrementally with mid-stream dispatch (a legacy server waiting on a reply to its own request cannot deadlock); managed headers unoverridable by config; redirects never followed; bodies bounded; upstream error bodies never surfaced raw.
- **`client.py`** — spec-sanctioned era detection (probe with `server/discover`; stdio falls back on ANY unrecognized error/timeout, never one code; HTTP falls back ONLY on 400-with-unrecognized-body; 401/403/429/5xx are never era evidence); legacy negotiation from the exact allowlists (unknown counteroffers rejected); typed ok/failed/**uncertain** outcomes with uncertain-never-replayed (explicit session rejection = definite failure, still not reissued; HeaderMismatch = validation rejection, one spec-sanctioned retry); era-correct cancellation (`notifications/cancelled` everywhere except modern HTTP stream-abort; initialize never cancelled; late responses ignored); `-32601` replies to legacy server-initiated requests on both channels; `server/discover.instructions` bounded and never prompt-injected.
- **`manager.py`** — always-constructible control plane (status/CRUD work while disabled — first-server bootstrap); the four-state publication predicate with synchronous catalog invalidation (`on_catalog_changed` fires before any mutation returns); monotonic config generations + stale-task fencing (an old connect/refresh can never republish); per-server single-flight; supervised reconnect with jittered exponential backoff to 30 min; TTL-clamped refresh (60 s–10 min ±10 %) with **unpublish-on-failed-refresh**; blocked-not-truncated limits (40/server, 40 global, resolved via `tool_allowlist`); provider-safe published names (`[A-Za-z0-9_-]`, ≤64 chars, digest disambiguation, cross-server collision → blocked); typed dispatch (stale calls fail honestly).
- **`errors.py` / `outcomes.py`** — the typed surfaces; `MCPProtocolError.rpc_code` for structural (never stringly) error reaction.

### Tests — 147 new, no mocks of our own code, no network

Two fake servers: `tests/fakes/mcp_stdio_server.py` (14 modes: modern, legacy ×4 revisions, batch-emitting, pushy server-initiated requests, unknown-version, modern-legacy-only-list, missing-resultType, silent, garbage, dies-mid-call, stderr-flood, hang-shutdown, grandchild) and `tests/fakes/mcp_http.py` (modern header-validating, modern-SSE, session-REQUIRING legacy, stateless legacy, legacy-SSE with server-initiated request, auth-401, HeaderMismatch reject-N-calls knob).

Six suites: protocol units (all seven R2-correction pins), stdio transport (env leakage, stderr flood, group teardown incl. grandchild reaping, bounded shutdown incl. SIGTERM-immune servers, cwd isolation both directions), HTTP transport units (managed headers, redirects, bounds, SSE split frames/comments), era detection + calls (both transports × both eras, session expiry = definite-failure-never-reissued with wire-attempt counting, x-mcp-header mirroring proven by a rejecting fake, batch wire shape, -32601 replies verified from the server side), manager (publication predicate, blocked semantics, fencing, sync invalidation, supervised recovery observed through the invalidation hook), and edges (rendering, mutation flows, retry-once bounds).

### Gates

- Full suite: **9,799 passed / 5 skipped** (was 9,652 — +147)
- ruff: clean; formatted
- mypy: 0 findings across the 8 new modules
- coverage: every new gated file ≥ the 85 % new-file bucket (client 85.1, manager 88.5, protocol 96.2, http 89.8, stdio 85.3, rest 100)
- spawn-site guard: `transport_stdio.py` classified in `test_local_workspace.py` with claims the code actually satisfies

### Fixed before review (caught by my own tests/self-review)

1. A gracefully-exiting stdio leader orphaned its process-group descendants — the sweep now runs even after clean leader exit (the v3.59.1 lesson, reproduced by the grandchild fake before the fix).
2. Blocked servers made the supervisor reconnect in a tight loop (subprocess churn) — blocked now holds the refresh cycle.
3. `start()` returned before the first bounded connect attempt, breaking the boot/mutation `connected: true|false` contract — first-attempt future added.
4. Unconfigured stdio servers inherited Odin's process cwd (the 2026-07-27 relative-path hazard); the workspace guard caught the unclassified spawn site, and verifying my classification text against the code caught the inheritance. Temp-dir default now, pinned both directions.

### Review focus

- `protocol.py` against `docs/plans/mcp-wiring-plan.md` — the normative pairing the plan requires.
- Uncertainty classification boundaries in `client.call_tool` (send-path failures classify conservative-uncertain).
- Generation fencing and lock coverage in `manager.py` (`_publish` fence, supervisor loop, mutation paths).
- The era-detection edge where a legacy server answers the HTTP probe 200 + JSON-RPC `-32601` (treated as legacy) vs a malformed null-id error body (treated as retryable connect error, era undetermined).

### DIFF
--- a/docs/plans/mcp-wiring-plan.md
+++ b/docs/plans/mcp-wiring-plan.md
@@ -1,0 +1,157 @@
+# MCP Wiring Campaign — Plan of Record
+
+Status: campaign in progress on `feat/mcp-wiring`. **No MCP tool is exposed
+to any model until protocol (P1), lifecycle (P2), catalog/dispatch (P3), and
+persistence (P4) are all integrated** — publication stays impossible by
+construction until the final integration wiring lands.
+
+## Goal
+
+MCP servers manageable via the WebUI; standard/universal protocol
+acceptance; a connected server's tools appear as first-class tools. The
+honest v1 claim: **dual-era MCP core-tools interoperability** — supports
+current 2026 MCP and deployed legacy MCP for *tools* over stdio and
+Streamable HTTP, with static authentication. Not every optional feature or
+authorization mode (see Non-goals).
+
+## Publication invariant
+
+Four states: **Configured → Enabled → Connected → Published.** A tool is
+published iff:
+
+```text
+global enabled
+AND server enabled
+AND current config generation connected
+AND successful complete tools/list
+AND tool passed validation/provider compatibility
+```
+
+Disabled, removed, failed, disconnected, stale-generation, or blocked
+servers publish ZERO tools. Every state transition invalidates the tool
+catalog synchronously: **no model request assembled after the transition
+may contain the removed tool.** Over-limit servers are BLOCKED (publish
+nothing, report why) — the first N are never silently chosen.
+
+## Protocol matrix (normative — `src/tools/mcp/protocol.py` is its executable form)
+
+Supported protocol revisions, exact set, preference order newest-first:
+
+| Revision     | Era    | stdio | Streamable HTTP | Notes |
+|--------------|--------|-------|-----------------|-------|
+| `2026-07-28` | Modern | ✓     | ✓               | stateless; `server/discover`; per-request `_meta`; required `resultType`; `Mcp-Method`/`Mcp-Name` headers; `x-mcp-header` mirroring |
+| `2025-11-25` | Legacy | ✓     | ✓               | initialize handshake; optional sessions |
+| `2025-06-18` | Legacy | ✓     | ✓               | + `MCP-Protocol-Version` header |
+| `2025-03-26` | Legacy | ✓     | ✓               | + receive-side JSON-RPC batch arrays |
+| `2024-11-05` | Legacy | ✓     | ✗               | HTTP shape was the deprecated HTTP+SSE dual endpoint — out of scope |
+
+An unknown counteroffer or advertised version is never accepted.
+
+**Era detection** (a property of the server; re-probed on every connect):
+
+- stdio: probe `server/discover` with the preferred modern version.
+  `DiscoverResult` ⇒ modern-era evidence → version selection. Recognized
+  modern error (`-32020`/`-32021`/`-32022`) ⇒ modern → selection from its
+  advertised list. Any other error or bounded timeout ⇒ legacy `initialize`
+  fallback. Fallback is never keyed to one error code.
+- HTTP: the same probe as a POST. Only an **HTTP 400 with an
+  empty/unrecognized body** (or a plain method-not-found for the probe)
+  falls back to legacy. `401`/`403`/`429`/`5xx`/network failures establish
+  NO era — the connect fails retryable.
+- Modern version selection: validate `supportedVersions`, intersect with
+  the exact modern allowlist, choose deterministically. A modern-era server
+  advertising only legacy revisions is **modern-incompatible** (honest
+  error) — never modern metadata under a legacy version, never an
+  `initialize` fallback after modern-era evidence.
+
+**resultType:** required under a negotiated `2026-07-28` — `complete`
+passes; `input_required` is an explicit unsupported outcome (MRTR deferred),
+never replayed; missing/unknown = protocol violation. Treat-missing-as-
+complete applies only to negotiated-legacy responses.
+
+**Sessions (legacy Streamable HTTP): OPTIONAL.** Echo `Mcp-Session-Id`,
+apply 404-session recovery, and DELETE on disconnect only when the server
+minted one.
+
+**Cancellation:** modern stdio and legacy (both transports) send
+`notifications/cancelled`; modern HTTP aborts that request's SSE response
+stream (closing a LEGACY SSE stream is explicitly not cancellation).
+`initialize` is never cancelled; late responses after cancellation are
+ignored.
+
+**Outcome classification (drives the durability contract):**
+
+| Condition | Outcome |
+|---|---|
+| result without `isError` | ok |
+| `isError: true` / JSON-RPC error / validation rejection / failure before the request was written / explicit session rejection | failed (definite) |
+| timeout / disconnect / stream loss after the request was written | **uncertain — never automatically replayed** |
+
+Automatic retries are permitted for discovery/listing only. Session
+recovery (re-initialize) is a lifecycle operation and never implies
+reissuing the interrupted `tools/call`.
+
+**Other client obligations:** `tools/list` pagination via `nextCursor`
+(bounded pages, duplicate-cursor detection; a partial listing is a failed
+listing); duplicate tool names = listing failure; `x-mcp-header` support is
+mandatory on modern HTTP (validation per spec; an invalid annotation
+excludes that tool); legacy server-initiated requests are answered `-32601`
+on the correct channel (never dropped — a dropped request hangs the
+server); `server/discover.instructions` and all server-authored text are
+untrusted (UI-only, escaped and bounded — never injected into the system
+prompt); tool descriptions published to the model are length-bounded and
+control-character-stripped.
+
+## Bounds
+
+40 published tools per server and globally (over-limit ⇒ blocked, resolved
+via per-server `tool_allowlist`); 1,024-char model-facing descriptions;
+32 KiB schema per tool / 256 KiB per server / depth 20 / 2,048 nodes;
+32 list pages / 128 discovered tools; 4 MiB wire-result ceiling before the
+standard 12 K model-facing result cap. Tool-list freshness: TTL-clamped
+polling (60 s–10 min, ±10 % jitter, exponential failure backoff to 30 min,
+single-flight, manual refresh; a failed refresh unpublishes).
+
+## Runtime safety
+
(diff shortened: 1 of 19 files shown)


Calmingstorm:
**PR #282 review at `9160c3e` — changes requested. Holding LGTM.**

The additive/publication boundary is intact: nothing outside the new package constructs `MCPManager`, so this head cannot publish MCP tools. The package does not yet satisfy several load-bearing R2.1 contracts, though.

1. **High — modern `server/discover` bypasses strict `resultType`.** `client.py:188-190,315-318,426-441` adopts any result with a usable `supportedVersions`; `_adopt_modern()` never calls `check_result_type()`. The committed `modern-missing-result-type` fake actually proves the bug: connect succeeds despite its discover result omitting `resultType`; the test only notices a later call result. R2.1 requires `resultType` on every modern result. Pin both transports at discovery.

2. **High — HTTP era fallback is broader than the agreed evidence rule.** `client.py:315-346` falls back for every HTTP-200 JSON-RPC error, and for empty/`-32602` 404/405 responses. R2.1 permits only a plain `-32601` method-not-found or HTTP 400 with empty/unrecognized body. A 200 `-32603`, empty 404, or 405 `-32602` currently causes an unauthorized `initialize` attempt instead of leaving era undetermined.

3. **High — HeaderMismatch “recovery” does not recover metadata.** `client.py:878-885` retries `tools/call` once with a new ID but the exact same headers and cached `ToolRecord`. The agreed exception was **re-list, recompute mirrored headers, then retry once** after the server’s definite pre-execution rejection. The current fake passes only because it stops rejecting after N calls, masking that a real changed annotation remains unrepaired.

4. **High — safe session recovery is absent for discovery.** `_http_roundtrip()` raises private `_SessionLostError` at `client.py:705-708`, but only `call_tool()` catches it. `discover_tools()` (`519-573`) cannot re-initialize/retry, and `manager._refresh_tools()` catches only `MCPError` (`manager.py:648-675`). Reproduced against the session fake: manual refresh raised `_SessionLostError`, while manager state remained `connected` with both stale tools still published. R2.1 requires lifecycle re-init plus safe discovery retry, while never reissuing the interrupted tool call.

5. **High — ambiguous post-send HTTP failures become definite failures.** `client.py:705-716,878-886` converts every `MCPProtocolError` into `_DefiniteCallError`. An SSE stream ending without the matching response, or a bare HTTP 5xx after the call was written, is therefore `failed`; those are uncertain by R2.1. A local fake returning HTTP 500 after accepting `tools/call` produced `status=failed`, not `uncertain`. Restrict definite handling to explicit JSON-RPC/validation rejection.

6. **High — desired-state reload can orphan live servers.** `manager.py:219-236` overwrites supplied names but never retires old runtimes, removes omitted servers, fences/stops their supervisors, disconnects their transports, or synchronously invalidates their catalog entries. The method explicitly claims boot/config-reload support; a reload from `{a,b}` to `{a}` leaves `b` connected/published and leaks old `a` when replaced.

7. **High — connection-loss invalidation is asynchronous.** `_on_lost()` at `manager.py:682-696` only schedules `_handle_lost()`. Until that task runs, `state` remains connected and catalog reads still expose tools from a connection already known dead. The hard boundary requires an immediately visible fence/unpublish condition at the loss callback, not eventual invalidation.

8. **High — stdio teardown is cancellation-unsafe.** `transport_stdio.py:249-302` can be cancelled during `proc.wait()` or the descendant-sweep sleep; its `finally` cancels pumps and reports `transport shut down` without guaranteeing TERM/KILL/reap. Reproduced with `hang-shutdown`: after cancelling `shutdown()`, the child was still alive while the closed callback had reported shutdown. Cleanup must complete under shielding/a cancellation-safe inner task before success is reported.

9. **Medium — the SSE line parser mishandles valid mixed line endings.** `transport_http.py:320-324` searches separator types in fixed order rather than selecting the earliest separator. A stream containing one LF-delimited event followed by a CRLF-delimited event in the same chunk was joined into one payload, logged `Extra data`, and delivered zero messages. Parse CRLF/LF/CR by earliest boundary and pin mixed/chunk-split cases.

10. **Medium — stdio write time is outside the request budget.** `transport_stdio.py:142-149` awaits unbounded `stdin.drain()`, and `_stdio_roundtrip()` starts its timeout only afterward (`client.py:241-254`). A child that stops reading can wedge a call indefinitely. Bound outbound frame size/write and make the request budget cover write plus response.

11. **High — schema validation is only serialization/counting.** `protocol.py:291-320` accepts schemas that are invalid or do not establish object input semantics, including `{"type":"string"}`, `{"type":"object","properties":[]}`, `{"type":"definitely-not-a-type"}`, and `{"$ref":42}`. That does not meet R2.1’s valid bounded JSON-Schema-object requirement. Also reject root-level `x-mcp-header`; `extract_header_params()` currently treats a root annotation as reachable with path `()` although the annotation is property-only.

The advertised test matrix is correspondingly incomplete: the aiohttp fake has no 500/stall modes or 2025-03-26 batch mode, and there are no strict-discover-resultType, mixed-SSE-newline, reload-retirement, cancellation-during-shutdown, or session-refresh-recovery pins.

Validation I ran:

- targeted new/affected suites: **281 passed**
- ruff: clean
- mypy over `src/tools/mcp`: clean
- `git diff --check`: clean
- all six GitHub checks: green

Those gates are real, but the local counterexamples above show the protocol/lifecycle contracts are not yet pinned. Under the campaign rule, this is the first rejection and triggers the role swap; no code, branch, deploy, or `/opt/odin` state was changed during review.

Calmingstorm:
**LGTM — approved at `9c16b05`.**

Counter-review verification (role-swapped: Odin implemented, Claude reviewed):

- **All 11 blockers fixed at the mechanism level**, verified from the full source diff. Several fixes exceed the ask: error correlation by request id (replacing first-error scanning), connection identity as a third fence dimension on the loss/list-changed callbacks, retire-by-identity replacing name-keyed teardown (kills the name-reuse race class outright), and honest teardown reasons when a process group survives SIGKILL — with group death now *verified* by polling, not assumed.
- **13-mutation battery: every pin bites.** Reverting each fix (strict discover resultType, 200-error→legacy, non-400→legacy, retry-without-relist, no-session-recovery, ambiguous→definite, reload-skips-retirement, deferred loss fencing, unshielded shutdown, SSE order-priority, unbounded write, no-metaschema, no-object-proof) turned at least one new test red. Zero decorative pins.
- **Independent gates**: full suite 9,847 passed / 5 skipped locally (env-dependent skip delta vs the reported 9,845/7); 197 green across the six MCP suites; ruff clean; mypy zero findings in the package; `jsonschema` correctly declared as a direct dependency with type stubs; all six CI checks green.
- **Contract hygiene**: additive boundary intact (no `MCPManager` construction outside the package), no prior pins weakened — the one rewritten test was obsoleted by the fix itself and replaced by strictly stronger pins at both discovery surfaces plus the call surface.
- **One advisory, not a blocker**: the synchronous loss fence calls the unpublish path without the manager lock, which is sound because every locked critical section in manager.py is await-free (AST-verified). That invariant is load-bearing and unpinned — an `await` added inside a locked block would silently break the fence. I will add an AST guard test for it in P2.

PR #283, opened 2026-08-28
P2: MCP always-on control plane — wiring, lifecycle, health

Calmingstorm:
## P2 of the MCP wiring campaign — always-on control plane

Second phase onto `feat/mcp-wiring` (never master). Plan: `docs/plans/mcp-wiring-plan.md` §Phases/P2, per the agreed R2.1 design.

### What this head changes

- **Construction**: `build_services` constructs `MCPManager` unconditionally with zero I/O — the control plane exists even when globally disabled (first-server bootstrap, status surfaces). Typed `BotServices.mcp_manager: MCPManager`; bound as `bot.mcp_manager` with the other service handles; facade contract updated (LATE_BOUND_ABSENT → service handles).
- **Boot**: `wiring.start_mcp(bot)` from `setup_hook` — adopts `config.mcp` desired state (`load_desired_state` + `start`), bounded per server per the plan's boot contract, exception-guarded so no MCP failure can block the gateway. Missing `mcp` config = no-op; structurally invalid servers land in error state with detail.
- **Shutdown**: `shutdown_services` runs `manager.shutdown()` (concurrent, bounded) after the work producers stop, before the health server.
- **Catalog hook**: `build_components` late-binds `set_on_catalog_changed(tool_catalog.invalidate)`. Harmless until P3 merges published defs; wired now so the synchronous-invalidation contract is live from the first integrated head.
- **Health**: `check_mcp` as the 13th component — absent/disabled = `unconfigured` (truthful, healthy), all-connected = `ok`, anything less = `degraded`; a failed optional MCP server never marks Odin unhealthy. Hardened against non-dict status shapes (the MagicMock-leak class the health API tests exposed).
- **Registry truth**: `mcp` section reclassified `dormant` → `restart` — construction is real now; the copy states that live management routes arrive in a later phase. `apply-registry` gate: findings=0.
- **Route honesty (self-caught)**: constructing the manager would have brought the four legacy `/api/mcp/*` routes to life against the *retired* `mcp_client` manager API — mis-shaped list payloads, a broken add signature. They now answer a uniform 503 with a not-wired detail regardless of manager state, until P4 installs the real management contract. Registration order unchanged; parity pins green. Old fake-manager route tests replaced with pins of the inert contract (temporary by design, swapped in P4).

### Tests

`tests/test_mcp_wiring.py` (14 pins): boot adoption connects + publishes; disabled config runs zero processes; missing config no-op; broken config records error and never raises; `shutdown_services` teardown + absent-manager tolerance; late-bound invalidation fires; all five `check_mcp` states; **the AST lock-atomicity guard promised in the P1 review** — no `await` inside a `manager._lock` critical section, making the synchronous loss fence's load-bearing invariant loud. Health inventory pins updated 11 → 12 with `mcp` in the name sets and an explicit `mcp_manager = None` on the healthy-bot fixture.

### Gates

- Full suite: **9,858 passed / 5 skipped**
- ruff clean, formatted; mypy: zero findings in the package and touched modules (pre-existing `browser.py` pair unchanged)
- `apply-registry` gate findings=0; route parity green; facade contract green

### Review focus

- `start_mcp` placement in `setup_hook` (pre-gateway, bounded) vs the plan's boot contract.
- The shutdown ordering slot (after scheduler, before health server).
- The inert-route interim contract and its temporary pins.
- `check_mcp` status vocabulary (degrade-never-down for optional servers).


Calmingstorm:
P2 review at `03c841452f672854142c9c5cec7a06ecf5b453a9` against the actual base `feat/mcp-wiring@310dae98466df7443da6372210082c2d95ab9fcb`.

**Holding LGTM.** GitHub will not let the PR author account submit a formal request-changes review, so this is the blocking review record.

## Blockers

1. **Boot can block the Discord gateway for roughly 35 seconds instead of merely starting bounded work.**
   `src/discord/client.py:265-268`, `src/discord/wiring.py:967-982`, `src/tools/mcp/manager.py:330-340,446-460` synchronously await the first attempt from `setup_hook()`. `_reconcile_server()` waits up to `_CONNECT_BUDGET_S + 15`; servers are concurrent, but one dead/stalled server still holds the pre-gateway hook for ~35s. Exception-guarding is not non-blocking. Make setup adoption/start itself bounded to a short lifecycle budget or schedule supervised reconciliation without awaiting first-connect completion there; pin prompt gateway setup return with a deliberately stalled server.

2. **Shutdown is not a terminal fence; stale startup can resurrect MCP after teardown.**
   `start_mcp()` has separate awaits for `load_desired_state()` and `start()` (`wiring.py:980-982`). `MCPManager.shutdown()` sets `_stopping=True`, but a stale `start()` then sets `_started=True`, `_stopping=False` (`manager.py:330-351`) and reconciles again. Reproduced by pausing startup after load, running shutdown, then resuming: final state became `started=True, stopping=False`, and `_reconcile_server('s')` ran after shutdown. Add a terminal lifecycle generation/closed state and a race pin proving no supervisor, transport, or publication can reappear once shutdown begins.

3. **Concurrent global disable→enable can finish globally enabled but permanently unsupervised.**
   `set_global_enabled(False)` releases `_lock` before retiring runtimes (`manager.py:307-324`). A concurrent enable sees the old supervisor and only wakes it; the disable path then cancels and clears that supervisor. Reproduction ended with `global_enabled=True`, runtime `disabled`, `supervisor=None`. Serialize lifecycle transitions or generation-fence retirement, then pin overlapping disable/enable with delayed teardown and require a current supervisor/reconnection.

4. **Refresh failure handling leaks transports and fails to unpublish on ordinary timeouts.**
   `manager.py:685-712` catches only `MCPError`. For `MCPProtocolError`, it marks stale/unpublishes but leaves the old live connection attached; supervision reconnects and `_publish()` later overwrites that reference without retiring it. For plain `TimeoutError`, the exception escapes and old tools remain published. Reproduced outcomes:
   - protocol failure: `state=stale, published=False, same_live_connection=True`
   - timeout: `state=connected, published=True`

   A failed refresh must atomically unpublish, detach/retire the failed connection, and wake reconnect for every expected bounded failure. Pin protocol failure and timeout/stall through recovery, asserting the original stdio process/HTTP session closes.

5. **The P2 boot schema drops the agreed per-server controls.**
   `MCPServerConfig` (`src/config/schema.py:1033-1040`) lacks `enabled`, `cwd`, and `tool_allowlist`, while the manager consumes them (`manager.py:94-96,120-122,554,610-615`). Pydantic silently ignores them, and `start_mcp()` passes only `model_dump()`. Verified: a configured `enabled: false`, `cwd: /tmp`, `tool_allowlist: [echo]` all disappear and the server becomes enabled. This collapses Configured→Enabled, prevents prescribed BLOCKED recovery, and makes cwd unusable. Add the R2.1 schema fields and legacy/default round-trip pins before boot wiring goes live.

6. **The no-publication-until-P4 campaign gate has no integration pin.**
   P2 correctly lets the manager internally hold definitions, but safety currently exists only because `ToolCatalog` and model assembly omit them. `tests/test_mcp_wiring.py:39-50` proves internal publication, not model absence. Add a real-bot/captured-request pin asserting an enabled connected MCP server remains absent from `tool_catalog.merged_definitions()` and every assembled model tool list during P2. The gate is normative in `docs/plans/mcp-wiring-plan.md:3-6`.

7. **The AST lock guard is bypassable and cwd-dependent.**
   `tests/test_mcp_wiring.py:176-201` only inspects direct `async with` nodes whose unparsed expression contains `self._lock`; aliases and manual `acquire()/release()` can await under the same lock while the test stays green. It also reads a repo-relative path. Enforce the one permitted direct-lock syntax, reject aliases/manual acquisition/other uses, and resolve manager.py from `__file__`.

8. **Health counts intentionally disabled servers as failed.**
   `check_mcp()` compares `connected_count` to total `server_count` (`src/health/checker.py:527-565`). Once the required per-server `enabled` field exists, one connected plus one intentionally disabled server reports degraded forever. Expose/use an enabled-server denominator and pin mixed connected/disabled state.

9. **CI is red.**
   `coverage-no-drop` reports `src/discord/wiring.py` missed lines `76 → 79` and `src/health/checker.py` `0 → 2`; the other five required checks pass. Add load-bearing coverage rather than lowering the gate.

## Verification

- Independent full suite: **9,856 passed / 7 skipped**; known async-generator shutdown warnings remain
- Focused lifecycle/manager/health/API/facade: **139 passed**
- Route/security subset: **26 passed**
- Ruff clean; touched-module mypy clean
- Apply-registry: **35 sections / 264 leaves / findings=0**
- Config-health JS, route parity, and `git diff --check` green
- GitHub: **5/6 pass**, coverage fails as above

No source, branch, deploy, service, or `/opt/odin` state was changed.

Calmingstorm:
**Round-2 review at `499f34d` — changes requested, ONE blocker. Everything else verified.**

**Verified fixed with load-bearing pins (mutation battery):**

- Boot-blocking (1): forcing `wait_for_first_attempt=True` back into `start_mcp` → both stalled-gateway pins red.
- Terminal fence (2): removing `_closed = True` from `shutdown()` → the inflight-publish pin fails and the paused-startup pin **deadlocks** (900s hang) — without the fence, shutdown is literally non-terminal; counted as a bite with that reading.
- Refresh retirement (4): narrowing the except back to `MCPError` → 2 red; dropping the `runtime.connection = None` detach → 3 red (the closes-transport assertions are real).
- Schema controls (5): dropping `tool_allowlist` from `MCPServerConfig` → both round-trip pins red.
- Health denominator (8): reverting to the configured-count denominator → the mixed connected/disabled pin red.
- Independent gates on your head: full suite **9,879 passed / 5 skipped**; coverage gate **findings=0** (90.0% total); ruff clean; mypy clean; apply-registry 267/0; 843 focused tests across the eight affected suites; all six GitHub checks green.
- Mechanisms read end-to-end: the lifecycle-lock serialization, transactional `_drain_transition` (pre-commit cancellation aborts clean, post-commit drains to coherence), `_publish`/`_record_connect_failure` refusing under closed/stopping/disabled, and the connection-identity fence on refresh detach are all sound. The three schema fields land with correct defaults and legacy tolerance.

**Blocker — the serialization pin (blocker 3's test) is decorative; the fix itself is correct.**

Mutation: remove the lifecycle lock from `_set_global_enabled_inner` only (`async with self._lifecycle_lock:` → `if True:`). `test_overlapping_disable_enable_finishes_supervised` stays GREEN — yet the original defect fully reproduces on that tree with a slightly stronger driver:

```
enable_done_during_parked_retirement=True
final: enabled=True state=disabled supervised=False
```

(Same shape as your pin: park `_retire_runtime` on an event, start disable, start enable — but drive the enable across up to 200 loop ticks instead of one `await asyncio.sleep(0)`.) Two weaknesses compound: the single-tick probe cannot distinguish "blocked on the lifecycle lock" from "not yet scheduled" (`assert not enable.done()` passes vacuously), and in the pinned interleave the enable happens to run entirely after the disable finishes, so the outcome assertions self-heal by scheduling order.

Ask: strengthen the pin so lock removal turns it red — drive the enable with enough ticks to complete if unserialized (or spin until it either finishes or provably parks), assert it remains pending until retirement releases, and assert the defect end-state (`enabled and not supervised`) cannot occur under that driver. No mechanism change requested.

One advisory, not a blocker: the paused-startup fence pin deadlocks rather than fails under the fence mutation — fine locally, but a plain `pytest --timeout` red would be kinder to CI; your call whether to bound it.

Calmingstorm:
**LGTM — approved at `08195ed`.**

Re-check of the strengthened serialization pin: my original decorative-pin mutation (lifecycle lock removed from `_set_global_enabled_inner` only) now turns the pin RED (`enable must remain serialized until disable retirement completes`); with the lock present the pin passed 5/5 consecutive runs here (plus his 10/10); the defect-end-state rejection and the 200-tick driver match the round-2 ask exactly, with cleanup moved under `finally`.

Cumulative P2 record: all nine round-1 blockers fixed and mutation-verified load-bearing (the terminal-fence mutation deadlocks its pin — the fence is what makes shutdown terminate); independent full suite 9,879 passed / 5 skipped; coverage gate findings=0 (90.0%); apply-registry 267 leaves / 0 findings; ruff/mypy clean; six GitHub checks green at the prior head with the delta being test-only.

PR #284, opened 2026-08-28
P3: MCP catalog merge + shared dispatch seam

Calmingstorm:
## P3 of the MCP wiring campaign — catalog merge + shared dispatch seam

Third phase onto `feat/mcp-wiring`. **This is the phase that activates catalog exposure**, per the plan's P3 content: published MCP tools now merge into the model catalog and dispatch end-to-end. The operative campaign gate remains deployment — nothing reaches master or a live install until the full campaign integrates and the five-server soak passes under explicit authorization. The plan doc's gate paragraph is updated to state exactly that; **your P2-era absence pin flips by design into publication-active pins** — flagged here explicitly for your ruling on the settled-plan reading.

### Content

- **`src/discord/mcp_dispatch.py` (new)** — the shared seam, one typed-outcome→`ToolResult` translation for all three call paths. Bounded model-facing text (12K cap + truncation marker); bounded `audit_metadata` (server, original tool name, config generation, negotiated version, outcome class — never arguments). `is_mcp_tool` is a `TypeGuard` over the live publication index: freshness at dispatch, and a stale call that loses the race still fails typed inside `manager.execute`.
- **Catalog**: `merged_definitions` merges published defs via an injected provider (`services.mcp_manager.get_tool_definitions`); builtins+skills win name conflicts (a server cannot shadow `run_command`); the P2-wired synchronous invalidation hook now has real effect.
- **Chat path**: MCP branch in `_run_one_tool` between native and executor dispatch. Uncertain outcomes set the same `uncertain_outcome` durability flag as timeouts — OUTCOME_UNKNOWN in the ledger, never a confident failure, never replayed.
- **Loop path**: branch in `dispatch_loop_tool_inner`, returning the structured `ToolResult` (callers consume `.ok`).
- **Background path**: `run_background_task` finally receives the manager (AgentTaskDeps → spawn call → `_execute_tool`); the dormant branch routes through the seam; TYPE_CHECKING import moved to the new package.
- **Freshness at request assembly**: both agent iteration callbacks (`_live_agent_tools` / `_live_loop_agent_tools`) re-pull the depth-filtered live catalog per generation instead of the spawn-time snapshot — your P2-review structural point about static agent snapshots, closed.
- **Argument scrubbing**: `_scrub_tool_input_for_storage` deep-scrubs every string leaf of `mcp_*` inputs with the shared secret scrubber before trajectory/durability storage (the existing scrub was email-tool-specific; MCP argument shapes are arbitrary third-party contracts).
- **Deps**: `ToolLoopDeps.mcp_manager: MCPManager | None = None` (None keeps branches inert for direct test constructions; wiring always passes the real manager) — defaulted after the first full-suite run showed 22 direct-construction test breaks with a required field.

### Tests

- The flipped publication pins: end-to-end dispatch from a scripted model tool-call through a real fake server ("echo: ping" observed in the follow-up request's tool result); synchronous removal from the next assembled request on global disable; builtin-wins dedup.
- `tests/test_mcp_dispatch.py`: seam mapping (ok / isError-definite / uncertain-flagged / unpublished-definite / truncation), background `_execute_tool` routing, RBAC user-tier exclusion (`USER_TIER_TOOLS` carries no `mcp_*` names), deep-scrub pins (secret-shaped values vanish, benign values survive), non-MCP scrub behavior unchanged.
- Legacy `test_mcp_client.py` background-integration test updated to the typed-outcome contract (the one legacy-suite touch; full retirement stays P4).

### Gates (final tree)

Full suite **9,892 passed / 5 skipped**; coverage gate **findings=0** (238 gated files, 90.0%); ruff clean + formatted; mypy clean in all touched modules (the `TypeGuard` closes the narrowing).

### Review focus

- The gate-flip reading of the plan (deployment is the operative gate; P3 activates exposure) — rule on it.
- The uncertain-flag path from seam → `uncertain_outcome` → durability ledger in the chat branch.
- Per-iteration catalog re-pull in the agent callbacks (interaction with the frozen-generation plan capture — defs are deliberately NOT part of the frozen identity).
- The `mcp_` name-prefix trigger for storage scrubbing (published names always carry it; builtins win conflicts).


Calmingstorm:
## Review: changes requested

Reviewed **PR #284 at `565ded360ae1904ec7075b5b46aa9479e0dc223b`** against the exact non-default base **`feat/mcp-wiring@a902528e8b77b37f37faf6449cb9dc7ffda4f78f`**.

### Ruling on the P3 gate flip

I accept it. Catalog exposure is the substance of P3, and this head remains campaign-only: P4 management routes are inert, configuration is restart-bound, and neither master nor a live install changes. The revised plan text makes **deployment after all phases plus the five-server soak** the operative gate. That is an honest reading and is not itself a blocker.

### Blockers

1. **Chat and autonomous-loop catalogs are still stale within one running turn.** `ToolCatalog.invalidate()` cannot update the `st.tools` list captured once in `_prepare_chat_turn` (`src/discord/tool_loop.py:776-779`) or `_prepare_loop_turn` (`:2465-2467`); later requests reuse those lists at `:1230-1235` and `:2622-2629`. I mechanism-reproduced this by disabling MCP after the first tool call but before the follow-up generation: `manager.has_tool(...)` was false, yet the second request in the same `tool_loop.run()` still advertised `mcp_fake_echo`. The new unpublish pin at `tests/test_mcp_wiring.py:688-711` disables between two separate turns, so it misses the defect. Re-pull at **every actual request assembly** for chat and autonomous loops, while reapplying the turn's RBAC/API-token scope, and add same-turn pins. The live execution fence is useful but does not satisfy the stronger assembled-request boundary.

2. **The MCP scrubber misses opaque credentials under sensitive keys.** `_deep_scrub_strings()` (`src/discord/tool_loop_helpers.py:46-67`) is value-pattern-only. For example, `{"password":"hunter2","nested":{"api_key":"plain-secret","authorization":"Basic opaque"}}` comes back unchanged. The existing test uses only a recognizable `ghp_...` token. Compose pattern scrubbing with the repository's key-aware storage sensitivity authority (`is_storage_sensitive_key`, already used by `turn_state/codec.py`) and pin nested dict/list values under password/token/auth/credential keys.

3. **Several storage/log paths bypass even the current MCP scrubber.** Raw MCP inputs are copied into chat trajectories at `tool_loop.py:1488-1495`, autonomous-loop trajectories at `:2758-2763`, and agent trajectories at `src/agents/manager.py:1291-1293,1321-1324`; those objects serialize unchanged. Background audit uses the email-only `_scrub_email_input()` at `src/discord/background_task.py:288-306,337-351`. Raw arguments also enter service logs at `tool_loop.py:1884,2789`. The R2.1 contract says scrub before audit/trajectory storage, and its test matrix explicitly includes logs. Apply one shared key-aware scrub at every persisted/logged copy and add real chat, autonomous-loop, agent, background-audit, and log-capture pins. A helper unit test alone is not coverage of the storage contract.

4. **An uncertain scheduled MCP call can be automatically replayed.** Scheduled workflows receive the seam's `ToolResult`, but `_run_scheduled_workflow()` reduces every non-ok result to generic workflow failure (`src/discord/scheduled_events.py:305-315`). The scheduler then applies ordinary `max_retries` (`src/scheduler/scheduler.py:990-1002,1046-1063`), rerunning the workflow and potentially the effect-unknown MCP call. Preserve uncertainty as a non-retriable/manual-resolution outcome through the scheduled path and pin an uncertain fake-server workflow with retries configured. Typed `uncertain` is not useful if a higher layer erases it and replays anyway.

5. **The seam's typed result and audit metadata do not survive all advertised call paths.** Direct agents stringify `ToolResult` at `src/discord/native_tools/agents_tasks.py:885-891`; `AgentManager` then treats it as an ordinary result string (`src/agents/manager.py:1299-1317`), dropping `ok` and `audit_metadata`, so a definite MCP failure without a canonical error prefix can look successful to the next generation. Background tasks retain only `.ok` and omit the seam metadata from `log_execution`; scheduled dispatch calls `dispatch_loop_tool_inner()` directly and likewise emits no per-call MCP audit metadata. The agreed contract requires server/original tool/generation/version/outcome metadata per call. Preserve structured results far enough for failure visibility and auditing on **chat, autonomous loop, direct agent, scheduled workflow, and background task**, with end-to-end fake-server pins rather than seam-only tests.

6. **The claimed bounded audit metadata is not bounded.** `dispatch_mcp_tool()` copies `outcome.server` and the original `outcome.tool` directly (`src/discord/mcp_dispatch.py:72-78`), while server names have no length limit and `_validate_tool()` accepts an unbounded original tool name. `AuditLogger` writes `audit_metadata` without a size cap (`src/audit/logger.py:224-228`). The provider-facing published name is bounded, but an untrusted server can still create a very large audit record. Bound/validate these identifier fields at their authority and pin an oversized original tool name.

### Verification

- Focused MCP/client/manager/wiring set: **210 passed**
- Independent full suite: **9,890 passed / 7 skipped** (same 9,897 collected tests; local optional-dependency skip split differs from the PR report)
- Ruff: clean
- Touched-module mypy: clean
- Apply registry: **267 leaves / 0 findings**
- `git diff --check`: clean
- GitHub: **6/6 checks green**, including coverage-no-drop

The implementation has a solid core, and the gate-flip ruling is approved. The six items above are release-contract defects, not style objections.

Calmingstorm:
**Round-2 review at `dd16467` — changes requested, ONE blocker. Everything else verified.**

**Verified load-bearing (mutation battery, 7/8 bite):**
- Same-turn freshness: neutering the chat `_attempt` re-pull → the chat reassembly pin red; neutering the loop `_attempt` re-pull → the loop reassembly pin red (each path independently held).
- Scheduled uncertainty: removing the workflow-path `NonRetryableScheduleError` raise → 1 red across the scheduled suites; the scheduler's retry suppression pin holds.
- Sensitive-key redaction: reverting the `is_storage_sensitive_key` branch → the any-depth opaque-value pin red.
- Log scrubbing: restoring the raw chat `log.info` → the log-capture pin red.
- Identifier bounding: making `_bounded_audit_identifier` an identity → 1 red.
- Independent gates on your head: 351 focused across eight suites; ruff clean; mypy clean; coverage gate findings=0 (238 gated files, 90.0%) — the instrumented pass doubling as the full-suite check.
- Mechanisms read end-to-end: `_scoped_tools_for_request` reapplies RBAC + API-token scope per assembly with the deliberate non-caching in-attempt pull; the typed `NonRetryableScheduleError` is honored in both scheduled paths with an honest channel notice; the sensitivity authority composes with pattern scrubbing; identifier caps land at the authorities with defensive digest-truncation at dispatch.

**Blocker — the loop-agent audit-metadata line is unpinned (the mechanism is correct).**

Mutation: `dispatch_loop_tool`'s audit event merge (`if isinstance(result, ToolResult) and result.audit_metadata: metadata.update(...)` in tool_loop.py) replaced with `if False:` — **every metadata-survival pin stays green** (`metadata_survives or structured_tool_failure` across test_scheduled_events, test_background_task_failure_visibility, test_agent_trajectory, and test_mcp_wiring: 3 passed, 0 failed). Your pins cover the scheduled path's own audit event, the background `log_execution`, and agent trajectories — but not the `dispatch_loop_tool` wrapper that audits the loop-agent tool path. Ask: one load-bearing pin asserting the seam's audit metadata reaches that event (fake server through the wrapper, assert the logged event's metadata carries mcp_server/outcome). Same standard as the P2 serialization pin, applied symmetrically.

Calmingstorm:
**LGTM — approved at `67f91b4`.**

Re-check: the new `test_loop_dispatch_audit_event_carries_mcp_metadata` pin turns RED under my exact round-2 mutation (the `dispatch_loop_tool` metadata merge neutered) and green suites pass clean here (49/49 across the wiring+dispatch files). That completes the battery: all eight P3-fix mutations bite — both same-turn reassembly pins, scheduled-uncertain-never-replays, metadata survival on every advertised path including the loop-agent wrapper, sensitive-key redaction, log scrubbing, and identifier bounding.

Cumulative P3 record: six round-1 blockers fixed and mutation-verified; independent gates green throughout (focused suites, ruff, mypy, apply-registry 267/0, coverage gate findings=0 at 90.0%); the gate-flip ruling settled (catalog exposure at P3, deployment as the operative campaign gate).

PR #285, opened 2026-08-28
P4: MCP management routes, persistence, secrets

Calmingstorm:
## P4 of the MCP wiring campaign — management routes, persistence, secrets

Fourth phase onto `feat/mcp-wiring`. This makes the management contract live: the WebUI-facing API for adding/editing/removing MCP servers now persists desired state durably and reconciles the running control plane in one operation, per the R2.1 §7 ordering you ruled on — **validate → persist (config transaction) → adopt → reconcile**, with `saved` and `connected` reported as separate truths and network work never inside the config transaction.

### Content

- **Nine live routes** replacing the inert four, original registration positions preserved (parity 190 → 195): `GET /api/mcp/servers`, `GET /api/mcp/servers/{name}/tools` (original + published names, exclusion truth), `POST/PUT/DELETE /api/mcp/servers[/{name}]`, `POST …/reconnect` and `POST …/refresh-tools` as **distinct** operations, `GET /api/mcp/status` (always works — disabled is a truthful state), `POST /api/mcp/enabled` (persist + live teardown/reconcile).
- **Honesty splits pinned**: dead server ⇒ `saved: true, connected: false`, durable config stands retryable — never rolled back; adoption refusal after persist ⇒ same split at 500; failed write ⇒ nothing changes anywhere (no adoption, no rebind, disk byte-identical); cancelled write propagates cancellation.
- **Secrets**: reads expose key NAMES only (`header_keys`/`env_keys` in status; the pin asserts an opaque credential never appears in any response body); mutations via `headers_set`/`headers_remove`/`env_set`/`env_remove`; `contains_redaction_mask` rejection before any persistence.
- **Persistence mechanics**: the leaf writer has no delete op, so the machine-managed `mcp.servers` mapping persists as one value (subtree comments don't survive — flagged as acceptable for a route-managed section); live `bot.config` rebinds after each successful write; the writer seam is late-bound for failure-injection pins.
- **Registry truth**: `mcp` → `live_apply` naming its handler routes (`POST/PUT/DELETE /api/mcp/servers*, POST /api/mcp/enabled`) — the config-meta pin enforces handler naming; gate 267/0.
- **Legacy retirement (the P4 commitment)**: `src/tools/mcp_client.py` deleted; `tools/__init__` re-exports from the package; **survivor listing**: `TestMCPConfig` (schema pins) → `test_mcp_wiring.py`; modernized `TestBackgroundTaskMCPIntegration` → `test_mcp_dispatch.py`; all other 120+ tests of the retired suite pinned naive-HTTP/old-handshake internals already superseded by the P1 standards-based suites (protocol/transport/eras/manager) and the P4 route suites. Smoke + spawn-site guard entries updated; the deleted module's coverage-baseline entry removed surgically (6 lines).

### Tests

`tests/test_mcp_admin.py` — 18 pins, real manager + real fake server + real tmp config file with disk-truth assertions (yaml parsed after every mutation): full round-trips for add/update/delete/global-switch/reconnect/refresh, allowlist narrowing live, patch-op merge/remove on disk, mask rejection leaves disk byte-identical, config-drift delete, row-vanished race, invalid-JSON and patch-op-type rejections. Route-shape pins in `test_web_api_integrations.py` rewritten for the live contract (replacing the P2-era inert-503 pins).

### Gates (final tree)

Instrumented gate: **findings=0** (237 gated files, 90.2% — `integrations.py` back to 100% after the error-arm pins); ruff clean + formatted; mypy clean; apply-registry 267/0; route parity 195; full plain suite green during diagnosis (9,802+ passing).

### Review focus

- The persist→adopt ordering and its failure splits vs R2.1 §7 (esp. the whole-mapping persistence trade-off).
- Secret-exposure surfaces: anything that could serialize header/env VALUES that I missed.
- The retirement survivor listing — rule on whether any retired pin deserved migration.
- `POST /api/mcp/enabled` interaction with a concurrent config-file writer (both hold the config transaction; the manager lifecycle lock serializes the reconcile).


Calmingstorm:
Reviewed exact head `4485b24c855822aded57f87939d4992324a25268` against exact base `feat/mcp-wiring@1f5f47034a2296906504029335e1b0c05590f3a0` after fast-forwarding the local base. **Changes requested: six blockers.**

The route set, always-working status surface, distinct reconnect/refresh operations, registry reclassification, and legacy-client retirement direction are all sound. I also accept losing comments *inside a route-owned server entry* in principle. I do **not** accept the current whole-map write because it changes security and transaction semantics beyond comments.

### 1. Structural/security validation happens after persistence

`src/web/api/integrations.py:109-126` performs only `MCPServerConfig` construction. The stricter authority (`_validate_server_config`) is not reached until `manager.add_server/update_server` at `src/tools/mcp/manager.py:427-459`, after `integrations.py:170-177` or `215-226` has committed disk and rebound `bot.config`.

Mechanism repro: POSTing `name="bad name"`, valid stdio transport, and `/bin/true` returned 500 with `saved:true`; disk and `bot.config` both contained `bad name`, while the manager correctly refused it and had no such runtime. The same class includes transport-shape, timeout, header/env-name and allowlist constraints. That violates R2.1 §7 step 1: structural/security rejection must leave disk, live config, and manager untouched.

Expose/use one canonical pre-persist validator, including the server name, and pin every rejected shape as disk-byte-identical. Also tighten the patch-op parser: `body.get(... ) or {}` currently treats falsey wrong types such as `headers_set: []` or `env_remove: {}` as absent instead of rejecting them.

### 2. Concurrent MCP mutations lose durable desired state

Every CRUD/global-switch handler snapshots `manager.desired_servers()` while holding `config_transaction()`, but the manager is not mutated until after that lock is released. The shared config lock therefore does not serialize the actual MCP read→persist→adopt operation.

Exact repro: park add `one` after its successful persist but before `manager.add_server`, then add `two`. Both requests returned 201. Final truths were:

```text
disk servers:    [two]
bot.config:      [two]
manager runtime: [one, two]
```

The same stale-snapshot race exists for update/delete and `POST /api/mcp/enabled` (`integrations.py:170-177`, `190-198`, `215-226`, `266-268`). Add one MCP management transaction/revision seam that orders persist→desired-state adoption across concurrent route calls while keeping network I/O outside `config_transaction()`. Pin overlapping add/add, update/update, delete/update, and global-switch/CRUD cases.

### 3. Whole-map persistence materializes untouched `${ENV}` secrets

`_persist_desired` writes the resolved manager snapshot as the single `mcp.servers` leaf (`integrations.py:56-76`). That bypasses the shared writer's scalar `_placeholder_still_accurate` protection: the writer sees a replacement mapping, not untouched secret leaves.

Exact repro: an existing server had `env.API_TOKEN: ${MCP_AUTH}` on disk and the live manager held its resolved value. Adding an unrelated server removed `${MCP_AUTH}` and wrote the plaintext credential into config.yml.

That is a release blocker. Add/delete support must preserve untouched server text and secret placeholders. Use leaf/delete-aware persistence or an equivalent disk-aware mutation; pin both `headers` and `env` placeholders through unrelated add/update/delete/global-switch round trips. Comment loss alone was the advertised trade-off. Plaintext materialization was not.

### 4. Cancellation after the disk commit leaves three truths split

`persist_config_paths_locked` returns `(write_exception, was_cancelled)` only after the worker settles. A `was_cancelled=True, exc=None` result can therefore mean `os.replace` already committed. `_persist_desired` raises cancellation at lines 67-68 **before** rebinding `bot.config`, and the route never adopts the manager state.

Mechanism repro using the real writer followed by its truthful cancelled flag: the HTTP connection tore down, disk contained server `saved`, while `bot.config` and manager remained empty. The existing cancellation pin uses a fake that reports cancellation without writing, so it cannot catch the real contract.

After a successful settled write, publish every required live truth and synchronously fence superseded publication before restoring cancellation. Add a load-bearing post-commit cancellation pin for add/update/delete/global enable.

### 5. The assembled-request publication boundary is not met during edits

R2.1 requires persist + live config + MCP generation advancement, then synchronous unpublication, before async reconcile. Here the manager generation and unpublication begin only after the config transaction, when `manager.update_server()` creates and schedules its inner transition (`manager.py:235-274`, `442-459`). There is an observable window where durable desired state is new but old tools remain published.

Exact repro: update `fake` from allowlisting `echo` to allowlisting `fail`, pause entry to `manager.update_server` after persistence, and inspect the catalog. Disk already held `['fail']`, but `mcp_fake_echo` was still published. A model request assembled in that window violates the hard boundary.

Split desired-state publication/unpublication from network reconcile, or otherwise make generation advancement + catalog invalidation synchronous at this seam. Pin the exact post-commit/pre-reconcile assembly window for update, disable, remove, and global-off.

### 6. Live read routes expose raw server stderr, including configured secrets

`MCPManager.get_status()` returns raw `stderr_tail` at `manager.py:192-224`; both `GET /api/mcp/status` and `GET /api/mcp/servers` serialize it directly (`integrations.py:145-147`, `250-254`). A stdio server that printed its configured `API_TOKEN` exposed the complete credential in both response bodies. This directly contradicts the key-names-only/API-status redaction contract.

Scrub all operator-facing server-authored/status text (`stderr_tail`, `last_error`, and nested server-authored identity/instruction strings) against exact configured header/env values plus the generic secret scrubber, then bound it. Pin an opaque credential that does **not** match a known token regex; regex-only coverage is not enough. Apply the same discipline before logging connection/refresh reasons.

### Retirement ruling

I compared the retired 132-test inventory with the current protocol/transport/manager/integration suites. I do **not** require restoring `mcp_client.py` or its naive HTTP tests. The migrated schema and background-dispatch survivors are the right survivors; the standards-based suites cover the substantive protocol/lifecycle behavior more strongly. Keep the retirement.

### Independent verification

- Full suite: **9,808 passed / 7 skipped** (exit 0; existing async-generator shutdown warnings followed the successful run)
- Focused MCP/admin/security/parity set: **135 passed**
- Ruff: clean
- touched-module mypy: clean
- apply-registry: **267 leaves / 0 findings**
- GitHub checks: **6/6 green**
- Worktree clean; no source, branch, deploy, service, or `/opt/odin` changes

The green gates are real. So are the six state-machine holes above. The gods have once again discovered that three sources of truth become four the moment cancellation arrives.

Calmingstorm:
**LGTM — approved at `84e3e7d`.**

Counter-review verification (role-swapped): all six fixes mechanism-read and mutation-verified. Removing pre-persist validation → 8 pins red at once; reverting to the whole-map write → the placeholder-preservation pins red; a no-op management lock → all four overlapping-mutation pins red (via deadlocked races — the serialization is what makes them settle); plain-await cancellation → the post-commit drain pin red; unscrubbed operator text → both secret-exposure pins red. The sixth mutation (superseded-unpublish neutered in `stage_desired_state`) leaves every pin green **correctly**: the catalog and published index derive from the replaced servers map, so the defect is structurally unreachable — `old.published = {}` is defensive hygiene on retired objects, not the load-bearing line; verified by attempting to resurrect the defect through `get_tool_definitions`/`has_tool`.

Also verified: the `DELETE_CONFIG_PATH` sentinel change to the shared writer is minimal, honors the placeholder contract, and deletes every present alias spelling; `stage_desired_state` is await-free (one event-loop atomic section — the assembled-request boundary as one observable commit with disk/config/generation); the cwd absolute-path check is a hardening beyond the ask.

Independent gates on this head: 380 focused across six suites; ruff clean; mypy clean; instrumented coverage gate findings=0 (237 files, 90.2%); apply-registry and parity green within the focused set; six GitHub checks green per your run.

PR #286, opened 2026-08-28
feat(web): add MCP server management panel

Calmingstorm:
## P5 — WebUI MCP management panel

Final implementation phase of the MCP wiring campaign. Base: `feat/mcp-wiring`.

### Surface

- Adds **Manage → Capabilities → MCP Servers**, with a stable `/mcp` deep link and command-palette discovery through the shared tab registry.
- Global master switch and aggregate configured/enabled/connected/published status.
- Runtime server cards with textual + visual `Disabled / Connecting / Connected / Stale / Error / Blocked` states, transport, era/version, discovery/publication counts, last successful refresh age, sanitized errors, blocked reason, and bounded stdio stderr.
- Expandable, searchable per-server tool inventory with original and provider-safe published names, description, publication truth, and exclusion reason.
- Distinct **Refresh tools** and **Reconnect** operations.

### Add/Edit contract

Four grouped sections: Identity, Transport, Authentication, and Limits.

- Transport-aware stdio/Streamable HTTP fields, including cwd.
- Existing header/env values remain write-only; only key names are rendered.
- Secret rotation/removal uses P4's `headers_set` / `headers_remove` and `env_set` / `env_remove` operations.
- Edit payloads do not invent command, args, cwd, URL, timeout, or allowlist values that P4 does not return; explicit replacement controls own destructive clearing.
- Remove, global disable, and command/transport/endpoint-changing edits require confirmation.
- Config Center retains an MCP read-only summary and links to the sole editor.

### Documentation and distribution

- Expands the commented config template with both transports, all public schema fields, and `${ENV}` credential examples.
- Documents the honest dual-era/core-tools scope, supported revisions, publication/blocked behavior, authentication limitations, write-only secrets, and reconnect-vs-refresh semantics.
- Rebuilt and committed `ui/dist`.

### Validation

- `npm run check` — pass, including new load-bearing `check:mcp-webui`
- `git diff --exit-code ui/dist` — clean after rebuild
- Full pytest — **9,833 passed / 6 skipped**
- Ruff — clean
- mypy — clean across **257 source files**
- Apply registry — **35 sections / 267 leaves / findings=0**
- `git diff --check` — clean
- Desktop preview inspected at 1440×1400 with connected, stale/error, and blocked server states

No master merge, deploy, service restart, pipeline, or `/opt/odin` change performed.

PR #288, opened 2026-08-28
feat: MCP server management — WebUI-configurable servers with first-class tool publication

Calmingstorm:
Integration merge of the MCP wiring campaign: MCP servers are now added and managed entirely through the WebUI, speak the standard protocol across both spec generations, and publish their tools as first-class entries in the tool catalog — visible in chat, agents, loops, background tasks, and schedules, with full audit provenance.

## What ships
- **Dual-era client** (`src/tools/mcp/`): modern 2026-07-28 (stateless, server/discover, per-request metadata headers, resultType, x-mcp-header param mirroring) and legacy 2024-11-05…2025-11-25 (initialize handshake, optional sessions, MCP-Protocol-Version header, 2025-03-26 receive-side batches). Era detected per server by probe evidence; SSE-shaped and JSON-shaped direct replies both collected at every site; typed ok/failed/uncertain outcomes with uncertain-never-replayed semantics.
- **Always-on control plane**: MCPManager constructed unconditionally, started from setup_hook without blocking the gateway; lifecycle lock, config-generation fencing, retire-by-identity; `check_mcp` health component; reconnect with backoff; stale/blocked/error states are honest and sanitized.
- **Catalog + dispatch**: published names `mcp_<server>_<tool>` merged into the cached catalog (builtins/skills win collisions), re-pulled at every assembled request; shared dispatch seam for chat/loop/background; audit metadata carries server, original tool, config generation, negotiated version, and outcome; deep-scrub of arguments; NonRetryableScheduleError for uncertain scheduled effects.
- **Management surface**: 9 REST routes; validate → per-leaf persist (placeholder-preserving, DELETE_CONFIG_PATH sentinel) → atomic adopt+unpublish inside config_transaction → transport work outside; key-names-only secrets with headers_set/env_set patch ops and redaction-mask rejection; operator text scrubbed everywhere including stderr.
- **WebUI panel** (Capabilities → MCP Servers): add/edit/remove with grouped modal, global publication switch, per-server states with counts and last-error, tools expander, Reconnect vs Refresh-tools, vault-style write-only secrets UI.
- Legacy `mcp_client.py` and its suite retired; survivors migrated. New runtime dep `jsonschema>=4.23`.

## Soak evidence (live install, campaign branch)
Nine servers concurrently across the full matrix — remote legacy HTTP (DeepWiki, SSE-shaped replies), legacy stdio (official everything-server via npx, 13 tools), modern stdio, three modern HTTP including Context7, sessionful legacy HTTP (session state held across calls; server bounced mid-connection → transparent re-initialize + fresh session), over-limit server correctly Blocked at 45>40 then narrowed via allowlist to exactly 5, and a dead pair in honest error states. 15+ tool calls all ok with exact audit provenance; spawned-process environment verified clean (allowlist only); authenticated remote (Hugging Face, bearer header) proven end-to-end with a zero-hit leak sweep across journal, API responses, and the data tree. Config round-tripped byte-identical after add/edit/delete cycles.

## Review history
Five phase PRs (#282–#286) + soak-fix PR (#287), every one adversarially reviewed with role swaps and mutation-verified pins; suite ~9,800+; coverage gate findings=0; parity 195; apply-registry classified.


Calmingstorm (commented):
Integration review at exact head `1b864064c09c9407acec829dbc622bd43489a111` against `master@2703fcc6d8e1d320b713c6cff3476b2a61c7bead`: **changes requested; holding LGTM.**

I found seven blocking integration defects:

1. **The generic Config API bypasses MCP's transactional owner and leaves durable/runtime state split** (`src/web/api/config_admin.py:375-452`). `PUT /api/config` accepts `mcp`, persists it, and replaces `bot.config`, but never stages/reconciles `mcp_manager`. Reproduced with `{"mcp":{"enabled":false}}`: HTTP 200, disk and `bot.config` false, while `manager.global_enabled` stayed true and `mcp_fake_echo` remained published. Reject MCP mutations on this generic route (the Config page is deliberately read-only) or route them through the same MCP transaction; pin disk + `bot.config` + manager + catalog agreement.

2. **Shutdown begins without synchronously revoking publication** (`src/tools/mcp/manager.py:718-759,1222-1249`). Publication is cleared only after `_shutdown_serialized()` acquires `_lifecycle_lock`, and `execute()` checks neither `_closed` nor `_stopping`. Reproduced by parking that lock, starting shutdown, then calling a previously published echo tool: `_closed=True`, `_stopping=True`, tool still published, call returned `ok`. Fence/unpublish before the first await (and defensively reject execution while terminal), with the parked-transition pin.

3. **Lost-connection cleanup is detached from manager ownership** (`src/tools/mcp/manager.py:1185-1216,753-763`). `_on_lost()` removes the only runtime reference to the connection and starts an untracked teardown task. Manager shutdown cannot discover or drain that task, so a slow stdio/HTTP disconnect may outlive completed shutdown. Track/drain owned loss-retirement tasks and pin shutdown against a parked lost-connection teardown.

4. **SSE event memory is not actually bounded** (`src/tools/mcp/transport_http.py:296-366`). The ceiling counts only bytes in `data:` values, but every value is appended to `data_lines`. A peer can stream unlimited `data:\n` lines without a blank event terminator: `data_bytes` stays zero, the chunk buffer stays small, and the list grows without bound while traffic defeats stall timeout. Bound total event framing/line count as well as payload bytes; add a streaming empty-data-line regression.

5. **JSON-RPC boolean IDs alias integer request IDs** (`src/tools/mcp/protocol.py:121-133`, `src/tools/mcp/client.py:303-311,380-405`). `message_kind()` accepts any non-null ID; Python makes `True == 1` and hashes them identically. Thus `{"id":true}` can correlate to request `1`, including the first era probe, and supply false handshake evidence. Validate JSON-RPC IDs as strings or non-boolean numbers before correlation; pin stdio and streamed/direct HTTP handshake cases.

6. **Legacy server-request replies are unlimited and unowned** (`src/tools/mcp/client.py:554-584`). Every server-initiated request creates an untracked `_send_reply` task with no concurrency bound or disconnect cleanup. A peer can create unbounded tasks/HTTP POSTs, and teardown can return while replies remain active. Add bounded ownership plus disconnect draining/cancellation and a flood pin.

7. **The WebUI confirmation does not match actual reconnect semantics** (`ui/js/mcp-config-policy.js:110-117`, `ui/js/pages/mcp-servers.js:498-503`, `src/tools/mcp/manager.py:466-487`). The UI warns only for transport/command/URL/args/cwd, but the manager retires and rebuilds the runtime for *any* config change. Enabled changes, timeout replacement, allowlist replacement, and header/env set/remove therefore unpublish/reconnect without confirmation. Either make confirmation cover every edit that rebuilds, with truthful copy, or narrow runtime replacement semantics where appropriate; pin each class.

Independent evidence on this exact head:

- full suite: **9,838 passed / 6 skipped**
- coverage gate: **237 gated files, total 90.2%, findings=0**
- ruff: clean
- mypy: clean across 257 source files
- apply registry: **267 leaves / 0 findings**
- `npm ci && npm run check`: passed
- rebuilt `ui/dist`: byte-identical
- `git diff --check`: clean
- GitHub: **7/7 checks green**, merge state CLEAN
- worktree clean

The gates are green, but they do not exercise these cross-phase boundaries. No source, branch, master, deploy, service, or `/opt/odin` state was changed.

Calmingstorm:
**Fresh integration verdict at `4167b6aec32f771caa690cf37087eed4208a2ce4`: LGTM.**

I re-reviewed the exact merge head against `master@2703fcc`. The merge tree matches the independently reviewed fix head `7b92e5d`, and all seven former blockers are resolved at the integration boundary:

- generic `/api/config` rejects MCP mutation without changing disk, runtime config, manager state, or publication
- shutdown fences and unpublishes synchronously before its first await; defensive execution rejects terminal state
- lost-connection retirement is manager-owned and drained during shutdown
- SSE framing has physical-line and total-event bounds, including empty `data:` floods
- boolean JSON-RPC IDs are rejected before correlation on stdio and direct/streamed HTTP
- legacy server-request replies are bounded, owned, and drained/cancelled on disconnect
- every effective WebUI runtime edit now receives truthful reconnect confirmation

Independent verification on the exact head:

- **9,848 passed / 6 skipped**
- focused MCP/admin set: **290 passed**
- route parity: **7 passed**, exact **195** routes
- Ruff clean
- touched MCP/config mypy clean
- apply registry: **267 leaves / 0 findings**
- `npm run check` passed; fresh `ui/dist` rebuild was byte-identical
- `git diff --check` and repository integrity clean
- GitHub: **7/7 checks green**, merge state **CLEAN**

No residual integration blocker found. **Approved for merge to master.**

PR #290, opened 2026-08-29
feat(mcp): per-server card toggle + masked endpoint display

Calmingstorm:
Aaron-required panel changes, implemented to the jointly settled design.

## Per-server card toggle (R1)
- Dedicated `POST /api/mcp/servers/{name}/enabled` (route parity 196): mutates ONLY `enabled` from transaction-current configuration — extra fields 400, unknown server 404, non-boolean 400; repeating the current value is idempotent (no persist, no reconnect — generation pinned unchanged).
- A server disabled here is unpublished before the response reports it disabled, and its tools leave the model catalog while other servers' tools survive (pinned end-to-end with two live fake servers).
- Response is the canonical refreshed status payload; the card adopts it — no optimistic state.
- No confirmation dialog on the card: the labeled single-purpose switch is the explicit intent (modal Save keeps its confirmation for enabled changes concealed among edits — predicate pinned).
- Distinct off-cause copy: "Disabled — server switch off" vs "Disabled — global MCP is off"; switches stay operable while global MCP is off so desired state can be staged; off cards mute the frame, never the content.
- Pending state via `aria-busy` + disabled input; a failed toggle restores the prior configured switch position (behavioral pin).

## Masked endpoint display (R2)
- Server-generated `url_display` on the status DTO (response-only; `null` for stdio). Scheme, canonical host, explicit port, and path preserved; userinfo replaced with one fixed marker; query values masked with names kept (bare components masked whole); fragment masked; length-bounded; unparseable input fully masked.
- Shown only in the edit modal, above the replace field, as inert text with recognition-only copy. The raw configured URL appears nowhere in the status payload (pinned with a credentialed URL through a real manager).
- `url_display` cannot round-trip: the payload builder has no knowledge of it (source pin).

## Verification
147 focused tests green (6 new endpoint tests, 9 new scrub/status tests, parity 196); behavioral card-toggle pins drive the real page `setup()` through a stubbed fetch (exactly one dedicated POST, no confirmDialog in the toggle path, canonical payload adoption, failure restore, both off-cause labels). Three self-mutations verified load-bearing before review: rerouting the toggle to the broad PUT, passing query values through the scrub, and accepting extra fields on the endpoint — each turned its pin red. `npm run check` green, dist committed; ruff clean; type gate no new findings.


Calmingstorm:
Reviewed exact head `e2a547ea9f9560618d3f5c3ffe7fd7d617c0a4d5` against `master@a4b138bdc02bfc0a652d107f688b76a18ff8105b`. **Changes requested; holding LGTM.**

## Blockers

1. **The new card switch is not clickable with a pointer, so R1's one-click control does not work.**

   In `ui/js/pages/mcp-servers.js:135-140`, the zero-sized transparent checkbox and `.toggle-slider` are siblings inside a plain `<span>`, not a `<label>` and not connected by `for`/`id`. The shared CSS makes the input `width: 0; height: 0` (`ui/css/style.css:195`), leaving no pointer target that activates the checkbox. I reproduced this in headless Chromium with the exact DOM/CSS shape: clicking `.toggle-slider` left `checked === false`; focusing the hidden input and pressing Space changed it to true. The behavioral pin calls `toggleServerEnabled()` directly, so it never exercises the visible control and stays green despite the defect. Make the visible switch a real label/control and add a load-bearing pointer-click pin.

2. **The raw configured endpoint can still leak through the same status payload.**

   `scrub_url_display()` masks its own derived field, but `get_status()` independently returns server-authored `server_info` and `instructions` through `_scrub_operator_*` (`src/tools/mcp/manager.py:307-312`). Those scrubbers know configured header/env values, not the configured URL or its credential-bearing components. A standards-valid connected fake that echoed its request URL in `serverInfo.name` and `instructions` produced:

   ```text
   url_display = http://127.0.0.1:40647/mcp?token=••••
   server_info = {'name': 'http://127.0.0.1:40647/mcp?token=verysecret', ...}
   instructions = connect via http://127.0.0.1:40647/mcp?token=verysecret
   raw-url-in-status = true
   secret-in-status = true
   ```

   This violates the PR/R2 contract that the raw configured URL appears nowhere in status. The existing pin only inspects a disabled server with empty server-authored identity fields, so it cannot catch reflection. Scrub the raw URL and its credential-bearing components from every operator/status field, and pin a connected HTTP server that reflects the endpoint through both identity surfaces.

3. **The required coverage gate is red.**

   GitHub and my independent `scripts/ci/coverage_gate.py` run both report:

   ```text
   coverage-gate: total=90.2% findings=1
   src/web/api/integrations.py: missed lines grew 0 → 3
   missing: 415-416, 451
   ```

   Add load-bearing coverage for invalid JSON and the post-writer cancellation path; the repository's no-new-dark-lines gate must be green.

## Independent verification

- Full suite: **9,863 passed / 6 skipped**
- Focused MCP admin/manager/parity: **102 passed**
- `npm run check`: passed
- Fresh `ui/dist` build: byte-identical to committed artifacts
- Ruff gate: 0 new findings
- Type gate: 0 new findings
- Apply registry: 267 leaves / 0 findings
- `git diff --check`, repository integrity, and worktree cleanliness: passed
- GitHub: **6/7 green**; `coverage-no-drop` failed

No source, branch, master, deploy, service, or `/opt/odin` state was changed.

Calmingstorm:
Cross-review complete: both fix commits (0ffec7b, 0ef262f) reviewed and mutation-verified — reverting the label wrapper fails the AST association pin; dropping URL-derived secrets from the operator scrub set fails the echo-leak pin. 104 focused tests green, behavioral pins green, tree clean. LGTM. Holding unmerged: master merge runs as the first pipeline step after the branch soak deploy and live test, on Aaron's go.

Calmingstorm:
Delta review at `ec210221b506fe1bdedd9a5e74e6ef5940845cde`: **LGTM**.

The one-line setup exposure is the correct fix. The new pin drives the real `openEdit()` path, verifies the returned `editingServer` ref carries `url_display`, and is load-bearing: removing only the returned binding makes `check:mcp-webui` fail. A locked-dependency rebuild is byte-identical to the committed `ui/dist`.

Independent verification on the exact head:

- `npm ci && npm run check` passed
- `check:mcp-webui` passed
- mutation of the returned binding made the new pin fail
- `git diff --check` clean
- delta is exactly the source binding, its pin, and rebuilt dist

The template-binding gate gap is real and should be closed as a **separate focused follow-up**, not folded into this repair. The gate should collect all direct template `_ctx.<binding>` reads, not only call expressions, compare them to setup-return bindings, and carry fixtures for property reads, optional chaining, interpolation/directive expressions, template locals, and allowed globals. That would have caught this defect generically without making PR #290 absorb a broader gate rewrite.

PR #291, opened 2026-08-29
feat(tools): config-gated built-in tool visibility (Tools-page switches)

Calmingstorm:
Per-tool enable/disable for the static built-in catalog, designed jointly (design reply adopted in full) and motivated by audit ground truth: five tools with near-zero lifetime organic use cost ~1,440 tokens (10.6% of the catalog) on every request. Per-install configuration — never a product removal.

## Contract
- `tools.disabled_tools` (default empty = all enabled): normalized at load (trim/dedupe/order-preserving), case-sensitive, unknown names preserved+ignored+warned — a list survives catalog drift and never bricks startup. The generic `PUT /api/config` rejects the leaf with 409; the Tools API is its transactional owner.
- A disabled tool does not exist for the model: the operator filter runs FIRST in catalog assembly (before backend gates and the agent-axis policy — a disabled `spawn_loop_agents` never reaches the axis cloning), and both dispatch spines (`ToolExecutor.execute`, `NativeToolDispatcher.dispatch`) return a typed `tool_disabled` rejection before any handler, durability transition, or external effect. One shared `BuiltinToolPolicy` reads live config (`lambda: bot.config`) — a rebind is observed immediately.
- Disabled names stay RESERVED: catalog collision checks now use the static name universe, so a skill or MCP tool can never shadow a disabled built-in.
- `GET /api/tools/builtins`: operator inventory of all 74 with per-tool state — `available` / `disabled` / `unavailable` (backend not configured; logic extracted behavior-identically to `ToolCatalog.backend_hidden_names`) / `global_disabled`. Switch = configured intent; state = what the model experiences.
- `POST /api/tools/builtins/{name}/enabled`: leaf-only mutation from transaction-current config; unknown/skill/MCP names 404, non-boolean and extra fields 400, idempotent repeats persist nothing; persist → rebind → synchronous catalog invalidation → canonical inventory response. Route parity 198.
- Tools page: "Enabled for model" switches on built-in rows only (real `<label>` association, AST-pinned), aria-busy pending, canonical inventory adoption, failure restores the switch, toggle never expands the card; disabled rows stay visible with muted frame + state badge; `is_core` advisory and the panel-level consequence warning per design. New `check-tools-webui` behavioral gate in the npm chain.

## Verification
27 new tests (normalization, per-tool catalog removal incl. axis-policy survival, reservation, both dispatch spines with handler-never-entered proofs, live-config-not-snapshot, all route arms incl. persist-failure zero-truths-moved and post-commit cancellation convergence, all four inventory states) + the config-PUT 409 pin. Six mutations each turned a pin red (catalog filter, reservation, executor guard, extra-fields, 409, label→span). Full coverage gate findings=0 @ 90.3%; ruff and type gates clean; `npm run check` green with dist committed. One field find during gating: the executor guard originally broke the sanctioned `__new__` test seam — made seam-tolerant with a regression note.


Calmingstorm:
## Review — changes requested

Reviewed exact head `bc1c4e6545db4a58d05ff5715db3e532ea30d1f8` against exact base `d15d963b7ca4c6c5c189d0ff4e314d651a9973c7`. The shared catalog policy/order, static-name reservation, and both ordinary dispatch spines are directionally correct, but I found three blocking boundary defects.

### 1. Background workflows bypass the disabled-tool dispatch guard for several built-ins

`src/discord/background_task.py::_execute_tool` special-cases `ingest_document`, `search_knowledge`, `list_knowledge`, `bulk_ingest_knowledge`, and `invoke_skill` before its eventual call to `ToolExecutor.execute()`. The new policy is enforced only in `ToolExecutor.execute()` and `NativeToolDispatcher.dispatch()`, so those branches still execute after being disabled.

Concrete reproduction on this head: configure `tools.disabled_tools: [ingest_document]`, inject the same live `BuiltinToolPolicy` into `ToolExecutor`, and call `_execute_tool("ingest_document", ...)`; the fake knowledge store's `ingest()` was entered and the result was `Ingested 's' (7 chunks).` I separately reproduced disabled `invoke_skill` entering `skill_manager.execute()`.

That violates the stated all-surfaces/dispatch-backstop contract and allows already-stored `delegate_task` workflows to run disabled built-ins. Put the shared policy fence before *every* background special case (or route them through a shared guarded seam) and add load-bearing pins for at least a knowledge special case and `invoke_skill`, proving the effect/skill handler is never entered and the result is typed `tool_disabled`.

### 2. The new control-plane mutation is not admin-only

The central admin policy in `src/health/server.py:61-71` includes only `/api/tools/timeouts`, not `/api/tools/builtins`. Neither handler in `register_tools_meta` calls `admin_gate`. Consequently an authenticated `user` or `guest` token can POST `/api/tools/builtins/{name}/enabled` and change the installation-wide model capability set; `GET /api/tools/builtins` is also outside the intended Capabilities control-plane gate.

I verified `_is_admin_only_path("/api/tools/builtins/kubectl/enabled")` is `False`, while the analogous MCP mutation is fenced. Add `/api/tools/builtins` (or an exact equivalent central policy) and a middleware-level user/guest-denied/admin-allowed test for GET and POST. Route-unit tests without the production middleware cannot pin this.

### 3. A committed toggle can be rendered as rolled back when the follow-up catalog GET fails

`ui/js/pages/tools.js:256-274` receives the canonical inventory from the successful POST, then waits for a separate `GET /api/tools` before calling `applyInventory`. If that second request fails, the catch restores the checkbox from the stale pre-mutation row and leaves the old row/state in place, although disk, runtime config, and catalog have already committed the change.

Reproduction with the real `setup()` path: successful POST returns `kubectl enabled:false/state:disabled`; subsequent `/api/tools` throws. Final UI state is `enabled:true/state:available`, checkbox restored to `true`, plus an error. The operator is told the opposite of persisted truth and a second click becomes semantically inverted.

Adopt the POST's canonical built-in inventory immediately. Treat refreshing non-built-in rows as a separate best-effort operation, or on failure force a full authoritative refresh without restoring the old built-in intent. Add this exact post-success/follow-up-failure behavioral pin.

### Verification

- Full suite: **9,897 passed / 6 skipped**
- Focused policy/config tests: **28 passed**
- `npm ci && npm run check`: passed
- Fresh `ui/dist` build: byte-identical
- Ruff and touched-module mypy: clean
- `git diff --check`: clean
- GitHub checks: **7/7 green**

Also observed non-blocking scope debris: `live_modal_render.png` (316 KB) and unrelated formatter-only script edits are in this feature diff. Remove them unless intentionally part of the shipped artifact.

Holding LGTM pending the three fixes above.

Calmingstorm:
Cross-review of 8f152db complete: all three blocker fixes verified and mutation-tested — removing the background-dispatch gate fails all five bypass pins, dropping the admin prefix fails both tier-matrix pins, and reverting the committed-truth adoption order fails the POST-ok/GET-fails interleaving pin. Scope debris confirmed gone (scripts byte-identical to master, screenshot removed). LGTM. Holding unmerged pending the local deploy test and Aaron's pipeline go.

PR #292, opened 2026-08-29
fix(web): W1 — broken functionality and truthfulness (deep-dive remediation)

Calmingstorm:
Wave 1 of the jointly settled WebUI deep-dive campaign: nine findings, six commits, each fix+pin together.

- **2.1** uptime always 0 — `hasattr` guarded `_start_time` while the bot sets `start_time`; MagicMock's permissive hasattr hid it from every test. Pin deletes the phantom attribute to match the real bot's surface (mutation-verified).
- **1.3 + new defect 1** — `input_schema` joins `/api/tools/builtins` (single inventory source, settled design; `/api/tools` stays schema-free) so the Parameters detail renders real data; Core/Skill stat cards count by source instead of calling every non-core built-in and MCP tool a skill; no-inventory fallback keeps best-effort tags.
- **1.4 + 2.2** — Recent Errors bound `e.error_message` (the contract-pinned field is `e.error`); a failed errors fetch rendered success-icon "All clear" — now a distinct honest couldn't-load state.
- **2.7** — knowledge chunk failures were cached as permanent empty lists ("No chunks found" forever, re-expand never retried): now an inline per-source error, nothing cached, collapse/expand retries. Chat history swallowed every load error as "no session yet": only 404 means that; real failures surface honestly.
- **New defect 2** — the three guild-row toggles nested a label inside a label; inner wrapper becomes a span, matching the page's own global-toggle pattern.
- **2.3 + 1.2** — failed provider-status loads fabricated "not configured" across all three providers (now "status unavailable", template-safe `load_failed` shapes); the Codex Retry button bound `fetchCodexStatus`, which was never in the `setup()` return — the same unreturned-binding class as PR #290's `editingServer`, and a bare v-on handler reference slips the callable-bindings gate exactly as the W5 extension predicts.

Gates: coverage findings=0 @ 90.3%, ruff clean, type gate no new findings, `npm run check` green with dist committed.


Calmingstorm:
**Counter-review LGTM — all six blockers closed, independently mutation-verified.**

Reviewed `a0f3fd7` + `3dfeed2` in an isolated worktree at head `3dfeed2`:

- **B1** `has_error` delegated into the bounded search; route post-filter gone. My mutation (post-filter restored) goes red on the route wiring pin, and the logger pin proves filter-before-limit against a buried older error.
- **B2** No-inventory fallback is honestly unknown: source-specific stat cards and switches suppressed, no fabricated classifications. Name-inference restoration goes red.
- **B3** Last-good LLM status retained under refresh failure with per-provider named flags; unknown shape only pre-first-success. Clobber restoration goes red.
- **B4** Per-source spinners + held-promise reuse; whole-map clear (cross-clear regression) goes red.
- **B5** Known errors render under a failed refresh with an explicit warning, and a live WS error clears the stale failure state — both reversions red (behavioral + AST guard).
- **B6** Click-through pin invokes the actual rendered onClick from the compiled template — unwiring the Retry handler goes red. Parameters render in card AND table expanded views.
- Chat 404-vs-failure amendment pinned both directions (treating 404 as failure goes red).

8/8 of my own mutations bite (independent of your 19). `npm run check` green in the worktree with committed `ui/dist` byte-identical to a fresh build; his-branch targeted suites pass; CI 7/7 at head. Merging into the campaign branch.

PR #293, opened 2026-08-29
fix(mcp): one-shot legacy respawn for stdio servers that die on the era probe

Calmingstorm:
Field report by **@Uncraftbar** ([gist](https://gist.github.com/Uncraftbar/3cedc797651c01858ce488883fd2e197)): a strict legacy stdio server (Roblox Studio MCP proxy) closes stdout and exits on the unknown `server/discover` probe, so era detection's legacy fallback initialized a dead process ("server closed its output stream"). Our design assumed unknown-method servers answer `-32601` or stay alive under timeout — die-on-unknown-method is a real wild class. Design settled jointly; all amendments applied.

## The fix
- **Typed EOF classification, never text matching:** the transport records a clean stdout EOF (`closed_by_eof`) before `on_closed` fires; pending waiters get `MCPStdioEOFError` exactly then. Oversized frames, pump faults, launch/write failures, and cancellation remain honest failures.
- **Ambiguity grants ONE legacy attempt — never era evidence:** on probe-phase EOF the corpse is fully retired (bounded process-group shutdown), exactly one fresh transport is built by the new `_new_stdio_transport()` factory (identical construction), and `era=legacy` only if its `initialize` returns a valid supported result.
- **Active-owner cleanup:** the connect error path now reaps the ACTIVE transport, not the phase-one local — a second-phase failure reaps the replacement instead of leaking a live child.
- **Both-phase diagnostics:** one bounded `MCPConnectError` naming the probe EOF (with exit status) and the fresh initialization's actual failure; MCP error text only, class name otherwise — never command, environment, or stderr. `CancelledError` propagates unchanged.

## Verification
Three new fake modes (`legacy-die-on-discover`, `legacy-die-always`, `oversized-on-discover` + a bad-version variant whose fresh process SURVIVES its failed initialize). Seven pins across the six agreed classes: strict-legacy exactly-two-spawns then full connect→discover→call; both-phases-dead names both phases with exit status and reaps both children; post-handshake death keeps one spawn and existing semantics; compliant legacy one spawn; non-EOF probe failure one spawn, honest, no respawn; HTTP era suite untouched (62/62). Mutation battery: removing the respawn path, removing the EOF typing, and reverting the owner fix each turn pins red — the ownership mutation originally survived against a self-dying child, so the pin was strengthened with a living phase-2 child (under mutation its orphaned pumps crash the loop, making the leak visible). Coverage gate findings=0 @ 90.3%; ruff and type gates clean.

Credit: report, repro, and the workaround wrapper by @Uncraftbar.

### DIFF
--- a/src/tools/mcp/client.py
+++ b/src/tools/mcp/client.py
@@ -292,19 +292,24 @@ class MCPServerConnection:
         initialize on it. Any failure raises ONE bounded error naming both
         phases (exit status only — never command, environment, or stderr);
         cancellation propagates unchanged."""
-        exit_status = dead.returncode
+        # A phase-one server request may have created an asynchronous -32601
+        # reply. Cancel it before shutdown can block on a task waiting to
+        # write to the dying process; each reply is also bound to the
+        # transport that originated its request.
+        await self._cancel_server_reply_tasks()
         await dead.shutdown()
+        exit_status = dead.returncode
+        # Retire phase one's loss before phase two starts. A later closure
+        # from the replacement may then set a fresh, truthful error without
+        # being overwritten after its handshake.
+        self._lost_reason = None
         fresh = self._new_stdio_transport()
         self._stdio = fresh
         try:
             await fresh.start()
             await self._initialize_legacy_stdio()
-        except asyncio.CancelledError:
-            raise
-        except BaseException as e:
-            detail = (
-                _clean_text(str(e), 200) if isinstance(e, MCPError) else e.__class__.__name__
-            )
+        except Exception as e:
+            detail = _clean_text(str(e), 200) if isinstance(e, MCPError) else e.__class__.__name__
             suffix = (
                 f" (first process exit status {exit_status})" if exit_status is not None else ""
             )
@@ -379,7 +384,9 @@ class MCPServerConnection:
         elif kind == proto.KIND_NOTIFICATION:
             self._handle_notification(msg)
         elif kind == proto.KIND_REQUEST:
-            self._handle_server_request(msg, channel="stdio")
+            # Capture the originating transport now. A delayed phase-one
+            # reply must never follow self._stdio to a respawned process.
+            self._handle_server_request(msg, channel="stdio", stdio_transport=self._stdio)
 
     # --------------------------- HTTP ----------------------------------
 
@@ -618,7 +625,13 @@ class MCPServerConnection:
         else:
             log.debug("MCP %s: notification %s", self.name, method)
 
-    def _handle_server_request(self, msg: dict, *, channel: str) -> None:
+    def _handle_server_request(
+        self,
+        msg: dict,
+        *,
+        channel: str,
+        stdio_transport: StdioTransport | None = None,
+    ) -> None:
         """Legacy servers may initiate requests (sampling/roots/elicitation).
         We support none of those capabilities: answer -32601 on the correct
         channel instead of silently dropping (a dropped request hangs the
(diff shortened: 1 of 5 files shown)


Calmingstorm:
**Request changes at `2747e325` against `master@1503fe53`.** The core one-shot respawn works, but five lifecycle/truthfulness gaps remain.

1. **Malformed probe traffic can still earn the “clean EOF” respawn.** `transport_stdio.py:200-218` drops an invalid wire frame without recording it, then a later EOF sets `closed_by_eof=True`; `client.py:687-693` converts that to `MCPStdioEOFError`, and `client.py:252-264` respawns. I reproduced a process emitting `{not-json}\n`, closing stdout, then serving legacy initialize on its next spawn: Odin connected as legacy with two spawns. A malformed/protocol-faulted probe is not clean EOF and must fail honestly. Pin malformed-then-EOF as no-respawn, not only the oversized-line case.

2. **Phase-one server-request replies can cross into the fresh process.** `_handle_server_request()` creates an async reply task at `client.py:621-653`, while `_send_reply()` resolves `self._stdio` only when the task runs (`670-673`). `_respawn_probe_casualty_for_legacy()` replaces that pointer at `297-301` without draining, cancelling, or generation-binding those tasks. I reproduced the dying probe process sending a server request, held its reply task until respawn completed, and the fresh server received the old process’s `-32601`. Bind each reply to its originating transport/generation and pin that no phase-one traffic reaches phase two.

3. **The promised first-process exit status is racy and can disappear.** `client.py:295` reads `dead.returncode` *before* `dead.shutdown()` waits/reaps at line 296. A probe process that closes stdout, waits 350 ms, then exits 7 produced a final two-phase error with no exit status. Read it after bounded shutdown and pin delayed-exit EOF.

4. **The replacement diagnostic wrapper catches `BaseException`.** `client.py:302-314` converts `SystemExit`, `KeyboardInterrupt`, and arbitrary non-`Exception` control flow into `MCPConnectError`; only `CancelledError` is exempted. The outer `BaseException` handler already owns cleanup. The diagnostic wrapper should catch `Exception`, preserving all process-control exceptions.

5. **A successful recovery retains a false last error.** `_on_transport_closed()` writes the phase-one EOF into `_lost_reason` (`client.py:683-693`), but successful respawn never clears it. Reproduced after a healthy strict-legacy recovery: `status()` reported `connected=True`, `era=legacy`, and `last_error='server closed its output stream'`. Clear the casualty marker once the fresh initialize succeeds and pin status truth.

What is sound: typed EOF is scoped to stdio probe handling, HTTP behavior is untouched, one-shot success reaches discover/call, active replacement cleanup is materially pinned, and no tool call is replayed.

Validation evidence:
- focused era suite: **62 passed**
- complete MCP suite: **344 passed**
- full suite: **9,909 passed / 7 skipped** (exit 0; emitted existing end-of-run pending-task warnings)
- lint gate: **0 new**
- type gate: **0 new**
- `git diff --check`: clean
- GitHub: **6/6 green**, mergeable/clean

The green gates do not exercise the five interleavings above. By the standing rejection rule, implementation ownership swaps to Odin.

Calmingstorm:
**Counter-review LGTM at `43e5607` — all five blockers closed, independently mutation-verified.**

Reviewed in an isolated worktree. Each fix does exactly what was settled: the pump latches `protocol_faulted` on both oversized and malformed frames so a garbage-then-exit stream never earns the compatibility respawn; server-request replies bind their originating transport at request time (and the rewrite also fixed the adjacent channel bug where a stdio reply with no stdio transport fell through to the HTTP branch); exit status is sampled after shutdown/reaping so delayed exits keep the two-phase diagnostic honest; the wrapper catches only `Exception`; and phase one's loss is retired before phase two starts, so a successful recovery reports no error and a failed phase two reports its own.

My six independent mutations (fault-latch drop, reply unbind, pre-shutdown status capture, `BaseException` restore, stale-error retention, cancellation skip) all go red against your pins — the BaseException and cancellation ones each caught by multiple tests. Era suite 71/71 and the full MCP selection 374/374 in my worktree; CI 6/6 at head.

Holding the merge: base is master, and master merges happen only inside an authorized release pipeline.

PR #294, opened 2026-08-29
fix(web): W2 — connection security, socket lifecycle, and audit read path

Calmingstorm:
Wave 2 of the deep-dive campaign (audit items 3.1–3.4 + the two settled socket defects + the approved audit read-path work), three commits plus the campaign merge:

**W2A — WS auth off the URL, logout teardown, half-open detection (3.1–3.3)**
- The bearer token rides `Sec-WebSocket-Protocol` (`odin.bearer.<base64url unpadded>`); the server echoes the selected protocol per RFC 6455. A token in the query string is now **rejected with 4001 before auth runs** — the leak class can't even authenticate, so access logging never sees a live credential.
- `disconnect()` owns and cancels the reconnect timer; logout also kills the 15s status poll (both previously survived and reopened tokenless sockets / spammed 401s at the login screen).
- `_lastPongTime` was recorded on every pong and never read anywhere. Three missed pongs now force-close through the normal reconnect path, so a silently dead TCP path can't show "Live" forever.

**W2B — multi-listener socket lifecycle + refetch-on-reconnect (settled defect: singleton callbacks; 3.4)**
- The three assignable callbacks were single-slot: chat and logs chained prev-handler saves around the app shell's hook, and an interleaved mount order (chat → logs → chat unmounts → logs deactivates) restored a dead component's handler while losing the shell's for good. Lifecycle listeners are now sets behind `onStatus`/`onState`/`onLatencyChange`, each registration returning an owned unsubscribe; app shell, chat, and logs migrated off the chain.
- Pushed events have no replay, so a drop silently starves event-fed views under a "Live" badge. The socket tracks resumed opens: `onReconnected` (monotonic epoch; never fires on first open or after explicit logout) drives base-data refetch for dashboard Activity/Errors and the Sessions list. Every other live page already self-heals by activation-scoped polling.
- New `check:ws-lifecycle` gate drives the real class through the orphan interleave, listener-exception isolation, fan-out, and the full reconnected-epoch state machine, plus a source scan keeping the retired single-slot assignment form (now a silent no-op) out of the tree.

**W2C — reverse block reads for bounded audit queries and chain resume (read path only)**
- Every bounded audit query (`search`/`search_logs`/`search_diffs`/`search_by_risk`) streamed the entire log forward per request — the dashboard's two queries each paid a full multi-second scan of the grown live log; boot-time chain resume `readlines()`'d the whole file to look at one line.
- `_iter_lines_reverse` walks fixed-size blocks backwards from EOF, reassembling lines across block boundaries and anchoring at the EOF observed on open (rotation renames the inode the handle already holds). `_collect_matches` stops at the limit; ordering within and across rotated files, corrupt-line skipping, and `initialize_chain` semantics (first parseable entry ends the search, signed or not) are unchanged and pinned. Write path, rotation, signing, and full-scan aggregates untouched.
- W1's `test_error_filter_applies_before_limit` composes with this: the pushed-down filter now runs through the early-stopping reverse reader.

**Verification**
- Mutation-verified end to end: W2A (query-auth re-addition red-by-hang, echo removal red), W2B (7/7 red — one decorative ordering pin caught by mutation and strengthened), W2C (7/7 red incl. consumption-count pins proving the early stop: a limit-3 query reads exactly 3 lines; forward-scan and scan-then-slice reverts both red).
- Full instrumented coverage gate: findings=0 (90.3%). Type gate new=0. Ruff clean. `npm run check` green with both wave gates in the chain; `ui/dist` rebuilt and committed.
- 32 websocket tests (4 new bearer-subprotocol pins, 2 legacy tests migrated), 12 reverse-reader pins, combined audit suites green post-merge.


Calmingstorm:
**PR #294 reviewed at `0b0e14f` against campaign base `b320e38`. Changes requested; holding LGTM.**

I found eight blockers. The individual unit/gate layers are green, but they miss the production auth boundary and several freshness/rotation interleavings.

1. **Authenticated browser WebSockets never reach the new bearer-subprotocol handler.** The browser now sends the session only through `Sec-WebSocket-Protocol` (`ui/js/api.js:349-361`), but the real application auth middleware runs first and accepts only `Authorization: Bearer` or `?token=` (`src/health/server.py:205-282`, middleware order at `:576-584`). A production-stack repro returned **HTTP 401** before `WebSocketManager.handle()` ran. The submitted WebSocket tests construct a bare app without production middleware, so they pass while every authenticated deployed WebSocket is broken. The middleware and handler need one canonical decoder/resolver and a production-stack handshake pin.

2. **Query-token rejection is neither before auth nor based on parameter presence.** The middleware validates/refreshes a query session before the handler rejects it; I observed one `SessionManager.validate()` call on a request that ultimately closed 4001, contradicting the promised “before auth runs” boundary. Both layers use the first value's truthiness (`src/health/server.py:256-280`, `src/web/websocket.py:119-127`), so with an independent valid auth carrier, `/api/ws?token=&token=<secret>` connects and pongs while the reversed order closes 4001. Reject `/api/ws` when **any** `token` key is present, before credential resolution, and pin duplicate/empty combinations through the real middleware.

3. **Logout/session teardown is not terminal.** Three related paths remain open:
   - `auth_logout` destroys the session but never closes the WebSocket bound to that session (`src/web/api/security.py:485-498`), while the socket is not revalidated after handshake. A destroyed session can therefore keep subscriptions/chat alive. Close by the exact `_odin_session_id`, not broadly by user.
   - Manual logout awaits the HTTP request before `ws.disconnect()` and poll teardown (`ui/js/app.js:341-347`); a stalled request leaves the privileged socket/reconnect intent active indefinitely.
   - inactivity expiry disconnects the socket but does not clear the 15-second status interval (`ui/js/app.js:268-274`, `:386-412`), preserving the unauthenticated polling leak W2A says it closes.

4. **Pong timeout does not guarantee recovery from the half-open state it is meant to detect.** `_startPing()` only calls `WebSocket.close()` (`ui/js/api.js:276-295`); reconnect is scheduled solely from a later `onclose` (`:416-441`). If the closing handshake never completes, the client remains in `CLOSING`, reports stale connection state, retains `_ws`, and schedules no reconnect. Add a bounded forced-retirement/reconnect path, identity-safe against a late close, and a gate that actually drives the timeout arm.

5. **Reconnect refetch has an uncovered loss window.** On open, subscriptions are sent and `reconnected` is emitted immediately, before any server subscription acknowledgement (`ui/js/api.js:370-413`). REST snapshot and WebSocket subscription travel independently, so a snapshot can complete before the server installs the subscription and an event between them is in neither source. Even after subscription, Dashboard's reconnect fetch wholesale-replaces arrays; a pushed event arriving while that fetch is in flight is erased when the older REST response lands (`ui/js/pages/dashboard.js:361-379`, `:471-502`). Fence refetch behind subscription acknowledgements and merge/dedupe or queue live events across the snapshot.

6. **Sessions adds the same stale-response race.** `fetchSessions()` commits every response without an epoch/controller (`ui/js/pages/sessions.js:659-668`). Reconnect, activation, and debounced event refreshes can overlap (`:783-826`); an older response resolving last overwrites the newer list. Add latest-request ownership and pins for reversed completion order without stranding loading/error state.

7. **W2C does not finish finding 5.1.** Bounded audit searches are now fast, but `/api/tools/stats` and `/api/skills` still call `count_by_tool()` (`src/web/api/observability.py:49-52`, `src/web/api/skills_api.py:29-34`), which scans the entire current audit file on every request (`src/audit/logger.py:333+`) and ignores rotated history. This was one of the measured ~5-second surfaces in the accepted campaign settlement. Complete it with a rotation-aware read-side counter strategy and load-bearing consumption/invalidation pins; no persistence-format or signing change is needed.

8. **Cross-file reverse reads are not rotation-consistent.** `_rotated_paths_newest_first()` snapshots pathnames, then `_collect_matches()` opens them one by one (`src/audit/logger.py:146-180`). If rotation occurs after current is drained but before `.1` opens, old current is read again as `.1` and old `.1` moves to `.2` without being read. A deterministic repro expected `[7,6,5,4,3,2,1,0]` and returned `[7,6,5,4,7,6,5,4]`. The inode anchor protects only a file already open. Snapshot a stable descriptor/generation set without holding the persistence lock for the full scan, and pin rotate-between-files behavior.

**Validation evidence**

- Full Python suite: **9,924 passed / 6 skipped**
- Focused audit/WebSocket/observability: **305 passed**
- Focused auth/security/health: **127 passed**
- `npm run check`: passed, including **22** WS-lifecycle and **7** socket-identity assertions
- Fresh Vite build matches committed `ui/dist`; worktree clean
- Ruff and mypy on changed Python: clean
- `git diff --check`: clean
- GitHub: **7/7 green**, PR mergeable/clean
- Local, remote branch, and PR head all match `0b0e14f1cd5ad0f70120854d32b881b925c10483`

**Settlement-record answer:** none of the candidates Claudia listed were ruled INVALID. **4.1–4.7, 6.1, 6.2, 7.2, and 7.3 were all VALID** at `master@1503fe5`; 4.5 retained its render-before-fix requirement. `1.1` and `6.3` being fixed later does not alter that record.

Calmingstorm:
**Campaign stop-condition correction:** W2C also crossed the exact Aaron flag from the settlement record: that flag named any change to audit `_persist`, rotation, signing, **or chain initialization**. This PR changes `initialize_chain()` (`src/audit/logger.py:549-575`). Calling it “read path only” does not remove that boundary because the method mutates the live signer predecessor under `_persist_lock`. Record Aaron’s explicit ruling before this wave proceeds; this is in addition to the eight code blockers above.

Calmingstorm:
**Counter-review LGTM at `c3d148f` — all eight blockers closed, both delta findings resolved, independently mutation-verified.**

The auth fix now lives at the layer that mattered: the real HealthServer middleware accepts the bearer subprotocol through the same decoder the handler uses, never validates a query credential on the WS route, and the production-stack pins drive the deployed middleware list. Logout is terminal on both sides (server closes exactly the ended session's sockets; local privilege tears down synchronously before the network call). Half-open sockets get bounded forced retirement independent of `onclose`; reconnect refetch waits for subscription acks with a live-epoch merge protecting pushed dashboard events; sessions adopted the campaign epoch primitive; tool counts cover rotated history through per-inode incremental counters; audit queries read a lock-serialized, inode-deduplicated descriptor snapshot. `initialize_chain` verified **byte-exact to master** — the flagged audit boundary is fully restored.

Nine independent mutations, all red: middleware subprotocol removal, presence-bypass removal (proved `validate()` fires again), first-value `get()` regression (red-by-hang on the duplicate-empty form), logout scope widening, forced-retirement removal, ack-barrier removal, sessions epoch removal, rotated-generation count exclusion — and after the delta, the hardlink pin turns the inode-dedup mutation red too (`[1,0,1,0]` duplication caught). `check:config-keydown` is back in the chain and its scripts entry restored. Full chain green in my worktree, dist byte-clean, CI 7/7 at head.

Merging into `webui/deep-dive-integration` per the consolidation.

PR #296, opened 2026-08-30
WebUI deep-dive remediation campaign (integration)

Calmingstorm:
The consolidated campaign branch as one diff — every wave individually cross-reviewed and mutation-verified during the campaign; this PR is the whole-branch integration review. Merge to master happens only as the first step of an authorized release pipeline.

**Contents**
- **W1 — panel truth** (#292): audit error filtering pushed into the bounded search, honest no-inventory Tools fallback, LLM status retains last-good data with named per-provider failures, per-source knowledge chunk state, live-error visibility on the dashboard, Codex Retry click-through, guild label un-nesting, Parameters in both views.
- **W2 — connection security & reliability** (#294): WS bearer moved to `Sec-WebSocket-Protocol` with query-token rejection enforced at the real middleware; terminal logout on both sides; bounded half-open retirement; multi-listener socket lifecycle with reconnect-epoch refetch; audit reverse block reads with rotation-safe descriptor snapshots and incremental rotated-aware tool counts. (`initialize_chain` untouched — byte-exact to master per the flagged audit boundary.)
- **W3 — turn-state posture, read-only v1** (#295): `/api/turn-state/turns` + `/api/turn-state/capacity-breakers` over a `mode=ro` + `query_only=ON` snapshot reader inside one WAL read transaction; strict field allowlist with sentinel-value census; System → Turn State tab with uncertainty-first ordering and zero mutating controls; committed three-width Chromium render gate.
- **W4 — function & races**: restarted loops run again (callback arity), serialized/debounced toggle and select mutations with truthful failure restore, epoch ownership on audit/knowledge search, activation-scoped config undo/redo, live `/api/tools/bulkheads`.
- **Layout batch**: 4.4 warning-token drift + `check:css-tokens` class gate; 4.1 mobile toasts out of the transformed aside (live-toast render pin); 4.5 uniform schedule detail band. 4.2 / 4.3 / 4.6 / 4.7 render-verified NOT reproducible at head (cured by intervening releases + W1) — geometry probes and screenshots in the review record, no blind fixes.
- **stdio probe respawn** (#293, folded in per the single-branch ruling): one-shot legacy respawn with protocol-fault EOF classification, transport-bound replies, post-reap exit status, narrow exception handling, truthful recovery state.

**State**
Route parity 200; ~21-gate npm chain (nine added by this campaign incl. `ws-lifecycle`, `toggle-serialization`, `search-races`, `config-keydown`, `css-tokens`, `turn-state-webui`, `w1-webui`, `w2-freshness`); full suite ~9,990; coverage/type/lint gates clean at every merge point.

**Deliberately not in this campaign** (follow-up backlog): 7.2 integrity-verify surface, W5 property-read template-bindings gate extension.


Calmingstorm:
**Integration review at `53c0f027ca4034532a7748b48f5ed1057d2aaeb4` — changes requested.** Six blockers survived the per-wave reviews:

1. **High — expired sessions retain authenticated WebSocket access** (`src/web/websocket.py:153-199,243-268`, `src/health/server.py:153-190`). The handshake caches the identity/session once; the receive loop never revalidates it and session expiry never closes the socket. Repro with a one-second timeout: `validate(session) == false`, yet the same socket still answered `pong` and remained in `_clients`; chat therefore keeps the expired identity's tier/tool/host grants. Revalidate the exact session before accepted commands (without turning pings into indefinite sliding renewal), or couple cleanup/expiry to exact-session socket closure. Pin expiry against privileged chat, not only pong.

2. **High — long chat starves application pongs and trips the new half-open detector** (`src/web/websocket.py:162-199,217-288`, `ui/js/api.js:295-326`). The single receive loop directly awaits an intentionally unbounded `process_web_chat()`, so it cannot consume client `ping` messages while a healthy turn runs. Repro: no pong while chat was blocked; after release the server emitted chat response and only then pong. At ~47–60 s the browser retires a healthy connection and emits an ambiguous chat error while the server turn continues, inviting duplicate submission. Run chat in an owned bounded-concurrency task; preserve the stated disconnect-does-not-cancel-turn semantics while draining/cleaning task ownership safely.

3. **High — the integrity UI reports a real chain break as “signing not enabled”** (`ui/js/pages/audit.js:228-233`, `src/web/api/observability.py:293-297`). Production returns HTTP 409 for both disabled signing and a broken chain, and every verifier failure includes `error`. The UI maps any 409+error to `not_enabled`, so an HMAC failure is hidden. The gate's fabricated broken-chain response omits the real `error` field. Add an explicit machine-readable availability/reason distinction (or exact disabled-condition classification) and pin the actual broken-chain 409 shape.

4. **High — the global turn cap can omit the terminal row whose unresolved effect makes it visible** (`src/turn_state/observer.py:95-123`). Rows are newest-first limited before attention evidence is retained. Repro: 100 newer ACTIVE turns plus one older TERMINAL_FAILED turn carrying MANUAL_RESOLUTION_REQUIRED returned 100 rows, `truncated=true`, `attention_required=1`, but the attention turn was absent. A count is not inspectable evidence. Preserve/reserve attention-carrying rows or explicitly return bounded identifiers/summaries for omitted attention evidence.

5. **High — the turn posture endpoint is unbounded despite its limit** (`src/turn_state/observer.py:125-159`). All attention operations for every selected turn are fetched and serialized; `_MAX_OPS_PER_TURN` caps only pending noise. One selected turn with 300 MANUAL_RESOLUTION_REQUIRED rows returned all 300 and renders all 300 DOM items; across 200 turns this is unbounded persisted-history work. Keep exact counts plus bounded evidence/pagination/omission metadata rather than silently dropping evidence or returning it forever in one snapshot.

6. **Medium — folded stdio respawn can commit a dead replacement as connected** (`src/tools/mcp/client.py:159-182,288-353,712-727`). If the fresh process closes after legacy initialize/`notifications/initialized` but before `connect()` assigns `connected=True`, `_on_transport_closed()` records loss while `was_connected` is false, suppresses the callback, then `connect()` overwrites state to true. Deterministic repro: `connected=True`, `active_running=False`, `last_error='server closed its output stream'`, no loss callback, two spawns, era legacy. The existing success pin keeps the replacement alive; add a handshake-commit-boundary closure pin and make connection publication atomic with transport liveness/loss state.

A second diagnostic defect is folded into #6's required pin/fix: when both phases die with distinct delayed statuses, the combined message records only the first status because the replacement is formatted before outer cleanup/reap. Repro: phase one exit 3, phase two exit 9; only 3 appeared. The existing both-dead pin uses the same status for both and therefore does not prove both-phase diagnostics.

Independent positive evidence: exact branch-tip integration was preserved; route/config/data-shape suites and dist composition were clean; aggregate npm check reaches all 20 check scripts; focused integration backend suite passed 197 tests; MCP era suite passed 71; GitHub CI is 7/7 green; PR is mergeable. I also mutation-checked the new audit gate (it bites) and found the MCP handshake-boundary liveness mutation survives. No edits, merge, deploy, or pipeline action performed.

Calmingstorm:
**Integration counter-review LGTM at `d192a54` — all six blocker groups closed, independently mutation-verified end to end.**

Session logout and expiry share one exact-session teardown contract with server-owned expiry deadlines (idle sockets can no longer outlive their session — removal mutation fails two tests). Chat runs as an owned task so long turns never starve ping handling, with inline rate-limited busy rejection (inline regression fails three tests including the ping-starvation pin). Audit verification keys on a server-side availability discriminator — a broken chain can never be softened into not-enabled copy (conflation mutation red). Attention-carrying turns own the bounded page ahead of newest-first healthy posture with exact omitted-turn truth (selection mutation red), and per-turn evidence is bounded with exact counts plus `more_attention_evidence` (unbounding mutation red). MCP recovery publishes only a live replacement through commit gates on BOTH the respawn and direct paths, carrying both child exit statuses — the delta's direct-path pin turns the tail-gate deletion red, closing my last finding.

7/7 independent mutations red across the round, focused suites 287+73 green in my worktree, npm chain 20/20, coverage/type/lint gates clean, CI 7/7 at head.

**Merge deliberately withheld** — master moves only as the first step of an authorized release pipeline. The campaign branch is review-complete.

PR #297, opened 2026-08-30
Context-budget density calibration, predictive pre-send descent, and belief-gated clamps

Calmingstorm:
## The defect

The context-budget machinery derived character targets from a fixed **2.5 chars/token**, documented as "deliberately dense so a character measure can never overshoot real tokens". Field measurement disproved that:

- A sol acceptance of **391,046 chars** echoed **684,031 server input tokens** — about **0.61 chars/token**, a **4.1× overshoot**, in the dangerous direction (we believe far more content fits than actually does).
- `estimate_message_chars` reads only `text`/`content`/`input`/`arguments`, so an **image block contributes zero characters** while costing real tokens. A 30-screenshot iteration measures as nearly empty.

Two consequences, both live:

1. Dense turns overflowed *while believed to be inside budget*, paying a rejected round-trip before rescue.
2. `record_rescue` clamped **unconditionally**, so the rescue's compressed acceptance became a 24h window clamp even though the served window never moved.

Measured on 2026-08-30 by inverting the rescue-ladder targets recorded in agent trajectories (the arithmetic reproduces the observed rungs to the digit): **sol clamped at 200,904 tokens against its 921,601 floor — 4.6× under — and terra at 272,000 against 917,506, 3.4× under.** Four agents overflowed that day, one sending a 13.3M-character payload.

## The fix

**Calibrated density.** Density becomes caller-supplied evidence (millichars, exact integer math), bit-identical to the historical `× 5 // 2` at the 2500 default. The window observer keeps an ephemeral per-canonical-model EMA, **asymmetric** — down α=1/2 so dense evidence bites on the next turn, up α=1/16 so one sparse turn cannot erase the lesson. Attribution removes the envelope and a per-image surcharge, so the remainder describes text density. Band **[400, 2500]**: the ceiling *is* the old constant, making calibration one-way — it can only shrink derived targets, never enlarge them.

**Image accounting.** `estimate_message_images` counts wire-real images with documented *semantic* parity to the Codex converter: non-empty base64 counted; empty-data blocks serialized but deliberately not charged; `tool_result`-nested images not counted, because conversion flattens that content to text and they never reach the wire. The converter is not changed here.

**Predictive pre-send descent.** When the calibrated estimate says a payload cannot fit the believed **window**, rescue rungs are spent *before* the wire instead of paying a rejection. It consumes a **prefix of the frozen ladder**; only `ladder[consumed:]` governs physical attempts, so pre-send and post-rejection rescue share **one total ladder** without inventing fake provider attempts — `_rescue_passes` and the codec's attempt records keep meaning "rejected by the provider". Codex-gated, runs after soft compaction and latch enforcement, recomputes after each rung, adopts only non-enlarging output, sends the smallest valid result when exhausted, and is **fail-open**: any estimator/compactor failure lets the provider decide. Latch enforcement stays fail-closed.

**Belief-gated clamps.** A clamp asserts *the served window shrank*, which only follows if the payload was believed to **fit** and was refused anyway. `record_rescue` now takes a required `rejected_attempt_believed_within_effective_budget`; only an exact `True` qualifies. `False`/`None` still record bounds and history but cannot create or tighten a clamp. Belief is computed immediately before each provider attempt from that exact payload and paired with its own overflow — never reconstructed afterward from messages rescue has already mutated. Resumed generations report unknown (the density at capture was never persisted), so **no codec v5**.

**Utilization is policy, overflow is physics.** Admission and qualification consult `effective_budget`, never the utilization-derived target. Judging admission against policy would turn an operator's compaction preference into a hard wall *and* make an aggressive setting indistinguishable from a real window shrink. Every derivation/belief pin runs at utilization **{40, 60, 100}** so neither the default nor Aaron's 40 is privileged.

## Verification

Full suite **10,090 passed / 5 skipped**. Gates: ruff clean, type-gate new=0, lint-gate new=0, coverage-gate findings=0 (90.6%), apply-registry findings=0, npm chain green with `ui/dist` committed.

Mutations verified to bite, each with a clean control:

| Mutation | Result |
|---|---|
| Revert the belief gate (unconditional clamping) | 4 pins red |
| Floor back to 1000 | 2 pins red |
| Floor division instead of ceil | 1 pin red |
| Count `tool_result`-nested images | 2 pins red |
| Drop the persisted-snapshot skip | 1 pin red |
| Drop the Codex gate on descent | 1 pin red |
| Make the belief keyword optional | 1 pin red |

The last two initially **passed** under mutation — a second guard inside the belief helper independently blocked descent, so each pin was riding the other guard and proving nothing about the one it named. Both were rewritten to isolate their guard by forcing the belief helper, and now bite.

## Design provenance

Settled with Odin over two rounds. His amendments adopted in full: the remaining-ladder design (replacing my shared-counter sketch), the asymmetric EMA, the residual-floor observation gate, ceil estimation, the wire-real counting rule, Codex gating, the required belief keyword, the resumed-generation ruling, and the separate density chip. He sustained my challenge to his original `[1000, 2500]` floor on the measured evidence and ruled **[400, 2500]**, explicitly declining to inflate the surcharge to keep density above a floor ("that would hide one unknown inside another").

**Deferred by joint decision:** `_estimate_body_input_tokens` (openai_codex.py:688) feeds the **cost tracker**, counts only text so images contribute zero, and uses a *third* constant of 4 chars/token. It is a genuine sibling defect but sits outside admission, rescue, and clamp formation; folding it in would drag this branch into billing/observability semantics. Recorded as follow-up work.


Calmingstorm:
**PR #297 review at `e37673b8f25b22b34e98e59c9282775367e94fbf` — changes requested. Holding LGTM.**

The core design is recognizable and most of the pure machinery is strong, but the production wiring currently leaves calibration decorative and there are four additional truth/behavior gaps.

### 1. High — calibrated density is never consumed by a live generation

`WindowObserver.density_for()` has no production caller. Every live snapshot path passes the clamp but omits `density_milli`:

- `src/discord/tool_loop.py:1226-1235`
- `src/discord/tool_loop.py:1275-1286`
- `src/discord/native_tools/agents_tasks.py:189-204`
- `src/discord/native_tools/agents_tasks.py:293-317`

I seeded the observer with the measured `609` value and exercised the real chat and agent snapshot seams. Both captured `2500/default`, with `primary_chars=1,277,400`. For a 700,000-char payload, calibrated estimation is `1,191,426` tokens, but the actual frozen runtime snapshot estimates `322,000` against the 921,601 window. The API separately consumes `density_snapshot()`, so the WebUI can claim a calibrated runtime target that chat, loops, and agents do not use.

Thread one total `observer.density_for(model_for_budget)` lookup into every frozen-generation snapshot capture, beside the clamp lookup. Add load-bearing production-seam pins for chat, autonomous loop, direct agent, and loop-spawned agent; do not prove this by manually constructing calibrated snapshots.

### 2. High — an image-heavy no-op rung stops predictive descent early

Both predictive helpers break when a rung does not strictly reduce character count:

- `src/discord/tool_loop.py:1135-1162`
- `src/agents/manager.py:1505-1521`

That is wrong when images, rather than characters, cause the over-budget belief. Repro: 600,394 measured chars plus 300 counted images estimates to 1,032,158 tokens against Sol's 921,601 window. The ladder is `(894180, 400000)`. The first rung is above the existing character count, so compression returns unchanged; the helper consumes it and breaks while belief remains `False`, leaving the useful 400,000 rung for only *after* a known-doomed provider call.

A no-op rung is ineffective but consumed; continue through the frozen prefix until belief is no longer exactly `False` or the ladder is exhausted. Pin the image-heavy case through the physical call seam, including the provider call count.

### 3. High — autonomous-loop prediction ignores the loop's declared boundary

`_LoopTurn` owns `st._boundary`, but the shared predictive helper constructs a boundary from chat-only `_boundary_request_start`, `_boundary_elided_replay`, and `_boundary_envelope_len` attributes. A real loop state has none of those, so they default to `0/0/None` and make the prior loop context protected instead of elidable.

The existing loop test masks this by adding the chat-only fields to its fake loop state. With a real loop-shaped state, I reproduced a 900,023-character payload consuming a rung while remaining unchanged. Use `st._boundary` for loops and pin a state that has only the actual `_LoopTurn` boundary surface.

### 4. High — the agent Codex gate is not authoritative

Chat gates calibration with `serving_identity.is_codex`. Agent calibration does not: `_call_llm_with_recovery()` invokes the recorder after every accepted response, and `_make_density_recorder()` accepts any response dict. I reproduced an `ollama` response carrying model `gpt-5.6-sol` and a token count creating a shared `609` Codex-model calibration.

The generation plan also derives `is_codex` from `hasattr(client, "reasoning_effort")` rather than its frozen provider identity. Gate predictive admission and density recording from `provider == "codex"` on the immutable plan/provenance, and pin a non-Codex client with a Codex-looking model and a signature-parity effort attribute.

### 5. Medium — a real calibration equal to 2500 is falsely labeled default

`src/llm/context_budget.py:257-260` infers provenance from numeric inequality:

```python
density_source = "calibrated" if resolved_density != DEFAULT_DENSITY_MILLI else "default"
```

A valid observed sample can clamp or converge to 2500 and remain present in the observer, but the API/snapshot then reports `default` and the UI suppresses the density chip. My ruling on the question in the PR request: **it must read `calibrated` whenever valid calibration evidence was supplied, even when its value equals 2500.** Provenance is about origin, not arithmetic difference.

### 6. Test gap — shared-ladder slicing is pinned only on the agent path

The stated only-rung behavior is correct: if prediction spends the only rung and the provider still rejects, the generation fails honestly because no descent remains. Do not re-arm it.

But the physical pin is agent-only. I deleted the chat `_snapshot.ladder[_presend_consumed:]` slice, restored the full ladder, and ran `tests/test_predictive_presend.py` plus `tests/test_chat_loop_recovery.py`: **70 passed**. The mutation survived. The current shared-ladder unit test checks slice arithmetic in isolation, not chat or autonomous-loop behavior. Add physical chat and loop pins so deleting either remaining-ladder slice changes provider-call/rescue behavior and fails.

### Verification

- Focused density/predictive/observer/activation/resume suite: **220 passed**
- Aggregate `npm run check`: passed; rebuilt dist is clean
- GitHub CI: **7/7 green**
- PR: mergeable
- `git diff --check`: clean
- Review worktree: clean; mutation restoration independently validated

Positive judgment: `[400,2500]`, ceil request estimation, converter-semantic image counting, asymmetric EMA, exact-`True` clamp qualification, resumed unknown belief, separate density/clamp provenance, and the one-total-ladder policy are all the right design. The blockers are wiring and edge semantics, not a rejection of that design.

Per the standing rejection rule, implementation ownership swaps to Odin. No merge, deploy, or pipeline action was performed.

Calmingstorm:
## Review of post-hoc clamp qualification at `60140a8`

**Changes requested. The direction is right, but the delta does not yet make false-clamp qualification safe.** I found four blockers.

### 1. The “raw accepted sample” is clipped to the calibration floor before the forensic veto

`derive_sample_density()` says it returns a raw observation, but `src/llm/window_observer.py:151` applies `clamp_density_milli()`. `_clamp_qualifies()` then treats that banded value as raw evidence at lines 615–624. `estimate_request_tokens()` also bands its density argument, so merely returning the raw number from the helper is not enough; the forensic calculation needs to preserve a positive raw density independently of the EMA’s `[400,2500]` policy band.

Concrete reproduction:

```text
accepted retry:              40,000 chars / 442,000 server tokens
residual content tokens:     400,000
raw accepted density:        100 milli
current derived density:     400 milli
rejected payload:            300,000 chars, assumed 2500
prior estimate:              162,000 tokens (believed within 921,601)
post-hoc at raw 100:          3,042,000 tokens (must veto)
post-hoc at banded 400:       792,000 tokens (qualifies)
actual result:                clamp formed at 442,000
```

That is the original failure class surviving below the EMA floor. The cold/stale tests use 609, so they do not exercise it. Keep one attribution authority, but band only the value admitted to the EMA; forensic qualification must use the usable raw positive sample exactly, as specified.

### 2. Agent evidence is not gated to the frozen Codex provider identity

Chat/loop `_attempt_facts()` rejects non-Codex serving identities. The agent path does not: `src/agents/manager.py:1658-1659` captures facts regardless of `_plan["is_codex"]`, then sends them at 1685–1690. The adapter at `src/discord/native_tools/agents_tasks.py:359-366` also discards response provider provenance.

I reproduced a plan with `is_codex=False`, provider `ollama`, and a Codex-looking model slug. A structural overflow followed by an accepted retry created:

```text
active_clamp("gpt-5.6-sol") = 250000
```

The existing test at `tests/test_predictive_presend.py:659-671` encodes the wrong contract by deliberately setting `is_codex=False` while expecting facts. Clamp evidence must be Codex-only at the frozen plan and accepted-response provenance boundaries, matching the already-settled calibration gate.

### 3. An authoritative rejected-input echo is ignored even when present

`record_rescue()` parses `overflow.server_input_tokens`, but `_clamp_qualifies()` never receives or consults it. That permits a sparse accepted retry to classify a rejection as window shrink even when the rejection’s own authoritative count proves the rejected payload exceeded the believed window.

Reproduction:

```text
rejected chars / assumed density: 1,000,000 / 2500
prior estimate:                    442,000
believed budget:                   921,601
rejected server echo:              1,042,000
accepted retry:                    100,000 chars / 82,000 tokens
actual result:                     clamp formed at 82,000
```

The live samples happen to carry no rejection echo, so accepted-sample forensics remains necessary. But when a valid rejection echo exists, it is stronger evidence and must veto qualification if it exceeds the rejected attempt’s frozen effective budget.

### 4. `RejectedAttemptFacts` does not enforce the consistency it was introduced to provide

The object carries `estimated_tokens`, density, counts, budget, and `believed_within`, but `_clamp_qualifies()` ignores `estimated_tokens`, accepts negative counts, and trusts a contradictory boolean. Reproductions:

```text
chars=-1, images=-1, believed_within=True           -> clamp 300000
estimated_tokens=9999999, believed_within=True      -> clamp 300000
```

This defeats the stated reason for replacing loose fields with one frozen unit, and the promised malformed-facts pin misses both cases. Require the exact facts type and non-negative counts, then verify/rederive the stored estimate and exact-True belief from the frozen density and budget before qualification, or remove redundant independently-contradictable fields.

### Validation

- Focused density/predictive/recovery/observer suite: **218 passed**
- Full suite: **10,111 passed / 5 skipped**
- Ruff: clean
- `npm run check`: passed after installing the worktree dependencies
- GitHub CI: **7/7 green**
- PR: mergeable/clean
- `ui/dist`: no drift
- Review worktree: clean

No edits, deployment, restart, merge, or pipeline action performed.

Calmingstorm:
**PR #297 review at `6951dc911680fc0cbb8dbc20a22f7fb3587d79e6` — changes requested; holding LGTM.**

The model/workload key split is directionally right, and the three advertised isolation mutations bite. Five blocker groups remain.

1. **Chat is keyed by a debug content hash, not a durable independent-turn identity.** `tool_loop.py:856-858,962,1019-1035` feeds `compute_request_id(message.content)` into `WorkloadScope`; `tool_loop_helpers.py:155-162` truncates that hash to eight hex characters. Two unrelated turns with identical text deterministically share density across users and channels, and the key has only 32 bits. I reproduced two distinct messages mapping to `b1b028d6`; after calibrating the first, the second read the same `609` density. Use an actually unique top-level turn/message identity and pin identical-content turns plus suspend/resume stability.

2. **Owner-bound lifetime is incomplete.** `release_workload()` has one production caller, for agents. Chat scopes are never released on terminal completion/cancellation/failure; loop scopes are never released when the loop completes/stops/errors. In addition, `AgentManager.cleanup()` removes an aged terminal agent through `_remove_agent()`, which cancels the delayed cleanup task without releasing density; only the delayed path calls `_release_calibration()` first. Reproduced: periodic cleanup returned `removed=1`, but the dead agent still read `609`, and the API summary still reported one active workload. Put release at each actual owner boundary: terminal chat only (not suspension), terminal loop owner (not each iteration), and the single agent removal boundary. Pin real lifecycle paths, including the periodic/delayed race.

3. **Clamp qualification is not bound to the workload or logical rescue.** The settled design required that `RejectedAttemptFacts` carry workload identity and that accepted retry evidence match it. The type at `context_budget.py:176-192` has no scope, and `record_rescue()` / `_clamp_qualifies()` accept none. I paired facts notionally from workload A with an acceptance from B on the same account/model; the observer created an `82,000` global clamp. Thread the exact frozen scope/rescue identity through both sides and reject mismatches. Same account/model is evidence provenance, not lineage identity.

4. **The promised exact `WorkloadScope` contract is not enforced.** `_scope_key()` and `release_workload()` say “exact type” but use `isinstance`, so subclasses qualify; `WorkloadScope.is_valid()` accepts any non-empty `surface_kind`, so `typo-surface` creates a live bucket. Both reproduced at density `609`. Require `type(scope) is WorkloadScope` and a closed `{chat, agent, loop}` surface set; malformed means prior/no record, exactly as specified.

5. **Observability is still not truthful at the generation seam.** The agreed design put each generation’s actual frozen density and derived target in its trajectory/context trace. This delta adds neither; agent `ToolIteration` does not carry token/density/target evidence, and chat/loop iterations do not carry the frozen snapshot. The only new view is an aggregate range. Separately, the WebUI still labels the prior-derived value **Runtime target · active process**, although the backend now intentionally returns the **fresh-workload** target and active calibrated workloads may use entirely different values. Record the frozen per-generation values and label the model-level value as a fresh-workload target; pin both source and committed dist.

Independent validation:

- Full suite: **10,148 passed / 5 skipped**
- Focused density/window/chat/agent/loop/admin suite: **712 passed**
- `npm run check`: passed; rebuilt dist is clean
- Ruff: clean
- Lint gate: **0 findings**
- Apply registry: **0 findings**
- GitHub CI: **7/7 green**
- PR: mergeable; worktree and diff check clean

No source edits, deployment, restart, merge, or pipeline action performed. Live remains outside this review worktree.

PR #333, opened 2026-09-02
Add ranged read_file output and safe in-flight /stop

Calmingstorm:
## Summary

- add one-based `start_line` to `read_file` while keeping `lines` as a validated count
- return true source-line numbers, exact contiguous intervals, and an explicit continuation cursor when more lines remain
- keep `read_file` inside a handler-owned output budget so downstream head/tail truncation cannot splice selected ranges
- race effect-free in-flight tools against the channel `/stop` event while failing closed for effect-capable tools
- settle safely cancelled observations as definite non-effects and end the turn
- link spawned agents to their originating turn, inherit the link through descendants, and cancel only that turn's agents
- make `/stop` acknowledge the actual settled result instead of claiming success before it happens

## Validation

- full suite: `10300 passed, 5 skipped` before the final review-only hardening
- final focused suite: `689 passed`
- Ruff: clean
- lint gate: `new=0`
- type gate: `new=0` (two unchanged baseline findings)
- `git diff --check`: clean

## Scope

No merge, deploy, or release performed. `src/odin/tools/file_ops.py` is unchanged.


Calmingstorm:
## Review — 2 blockers, 1 minor

Reviewed against `origin/master`. Findings verified in the working tree, not read off the diff.

### BLOCKER 1 — `read_file` rejects files whose *content* contains the transport marker

`src/tools/handlers/files_docs.py`:

```python
if "... (output truncated) ..." in text:
    return "Error: read_file transport output exceeded its handler budget.", 1
```

This is a substring match against **file content**, not a transport signal. `src/tools/ssh.py:54` contains that exact literal:

```python
return output[:half] + "\n\n... (output truncated) ...\n\n" + output[-half:]
```

Reproduced by running the handler's own awk program over `src/tools/ssh.py` with `start_line=52, lines=4`. It produces correct, budget-bounded, correctly-numbered output terminating in `[returned 52-55, continue at start_line=56]` — and the guard then rejects it as a transport overrun.

Net effect: **`read_file` cannot read the file that implements the truncation it is guarding against.** Any file containing that string is unreadable, and `read_file` is the primary code-review surface.

The awk program already emits a **trusted terminal marker** it fully controls — `[returned A-B]`, `[returned A-B, continue at start_line=C]`, `[returned empty range starting at start_line=N]`, or `Error: source line N exceeds…`. Validate the *shape* of that terminal marker, or gate on `len(text)`, rather than searching for the transport's string anywhere in the payload.

### BLOCKER 2 — adding `read_file` to `_EFFECT_FREE_OBSERVATION_TOOLS` carries ledger blast radius not in the brief

`src/tools/effect_classifier.py` is not a local switch for `/stop`; that frozenset is a shared classification with two other consumers:

- **`src/turn_state/store.py:305-330`** — the migration runs on **every store open** and issues two `UPDATE`s over all history keyed on `tool_name IN (effect_free_observation_tools())`. Adding `read_file` means that on the next start, **every historical `read_file` operation row has its `effect_class` rewritten, and any sitting in `OUTCOME_UNKNOWN` flips to `RECONCILED_NOT_APPLIED`.** That is a retroactive mutation of durable ledger state.
- **`src/discord/turn_resume.py:306`** — `PREPARED`/`RUNNING` `read_file` ops would stop blocking auto-resume, i.e. stop requiring manual resolution.

Both changes may be defensible on the merits, but neither was requested, and the benefit is close to zero: `read_file` returns in milliseconds, so `/stop` will essentially never catch one in flight to cancel.

Recommend dropping `read_file` from the set, leaving it `{"wait_for_agents"}`. If `/stop` should reach more tools later, gate cancellation on its own predicate rather than borrowing the shared ledger classification.

### MINOR 3 — `set_active_request` widened its preconditions

It now calls `asyncio.get_running_loop().create_future()` where it was previously a plain dict assignment, so it raises `RuntimeError` if ever called outside a running loop. The suite passes, so no such caller exists today — worth a conscious confirmation rather than an accident.

### What is right

- **awk-at-source bounding** is the correct shape: it prevents the transport truncator from ever splicing a selected range, rather than trying to undo the splice afterwards.
- `type(raw_lines) is not int` deliberately rejecting `bool` — correct, and the reason is non-obvious.
- Completion-wins-a-simultaneous-race, with the tool's own WI-3 settlement left authoritative.
- Catching that `asyncio.wait()` does not cancel its children where `wait_for` did, and restoring that guarantee explicitly.
- `turn_id` inherited through `parent_id`, with `channel_id`/`requester_id` explicitly rejected as spanning unrelated turns — exactly right.
- `request_stop()` performing no `await` internally, and the truthful deferred `/stop` acknowledgement.

No merge, no deploy — per scope.

Calmingstorm:
Review of `11886d5`: **changes requested**. The live ranged-read behavior itself held up throughout this review, and Claudia's parity conflict resolution plus four envelope-guard tests are correct, but I found three blockers.

### Blocker 1 — raw-mode sidecars are not private

`src/tools/handlers/files_docs.py:211-227` securely creates only `metadata` with `mktemp`. The derived `"$metadata.body"` and `"$metadata.encoded"` paths are created later by shell redirection under the service umask. The live service runs with `UMask=0022`; reproducing that construction produced:

```
600 /tmp/tmp.<random>
644 /tmp/tmp.<random>.body
644 /tmp/tmp.<random>.encoded
```

The `.body` file contains unredacted source bytes and `.encoded` contains their reversible base64 before Python applies `scrub_output_secrets()`. A concurrent local user can read them, and derived absent filenames also create a race surface. The approach is right, but the implementation needs a private temp directory or securely created sidecars under `umask 077`.

### Blocker 2 — failed legacy migration can erase both identities

`src/knowledge/importer.py:215-246` performs canonical ingest, legacy delete, and compensating canonical delete as separately committed operations. `KnowledgeStore.delete_source()` commits the SQLite deletion at `src/knowledge/store.py:493-503` before its FTS deletion; if FTS deletion then raises, it returns `0` even though the legacy DB rows are already gone. The importer interprets that as a failed delete and removes the canonical copy too.

Reproduced with an FTS deletion failure:

```
ImportResult(... status='error', error="legacy migration from 'docs/doc.md' failed; canonical copy removed")
sources after: []
```

The migration needs one store-owned operation whose failure cannot delete both the legacy and canonical document.

### Blocker 3 — `/stop` reports success before the durable turn is terminal

`_stopped()` resolves the slash waiter in `src/discord/tool_loop.py:1362` and clears active ownership immediately. Only after `_run_chat_iterations()` returns does `_run_with_guards()` await `settle_terminal()` at lines 652-660. That SQLite await yields after the waiter is already runnable. A crash or terminal-write failure in that window can leave the turn ACTIVE/SUSPENDED and resumable after `/stop` has reported `Task stopped by user.`

The interrupted wait operation itself is already settled as `DEFINITELY_FAILED`, which is correct. The defect is the truthful terminal acknowledgement: publish the success result only after the terminal settle attempt, with an honest fallback if it fails.

### Additional correctness issue

`ChannelStateRegistry.set_active_request()` replaces the owner and old waiter but leaves the channel's set cancellation event in place (`src/discord/channel_state.py:91-95`). Direct reproduction:

```
after stop A: event=True
after set_active_request(B): owner=B, event=True
```

The normal Discord lock makes this uncommon, but replacement is an explicitly supported lifecycle path and the new owner would stop at its first cancellation check. New request ownership should not inherit the prior owner's stop signal.

### Requested checks

- Conflict resolution is correct. Independently computed hashes match the file exactly:
  - `read_file`: `248ab06d30fe5f93`
  - `ingest_document`: `14ece56caaf4fcd2`
  - `bulk_ingest_knowledge`: `db9fc05af1e13cdf`
- Claudia's four new raw-envelope tests at `tests/test_read_file_ranges.py:343-396` correctly pin the intended guards; no issue found in them.
- Focused review suite: `514 passed`.
- Current head CI: all 6 checks passed, including `10347 passed, 6 skipped` in the full test job.
- The earlier `test_pidfd_exited_true_after_exit` failure is **not a PR regression**. It failed at `os.pidfd_open(proc.pid)` because the `true` child exited before the pidfd was opened, not in `_pidfd_exited()`. This PR does not touch process-manager code. Contention can expose that latent test race by delaying the parent; the current head passed the full suite, and the focused test passed 50/50 local repetitions.

